PR #8, opened 2020-09-24
FIX: rowActions return null shouldnt render bubble

ynotdraw:
### Changes

- TableModule `rowActions` can sometimes return `null`.

An example may be like in the story for this PR, where based on the `row` returned, you want to optionally show actions. 

```jsx
<TableModule
  rowActions={(row) => {
    // Only show actions when carbs are 37 or 67, otherwise,
    // we don't want any actionable buttons displayed for this row
    if (row.carbs === '37' || row.carbs === '67') {
      return actions; // your action buttons go here
    }

    return null;
  }}
/>
```

Currently, on master, the above example would yield you an empty toolbar:

94025892-04227200-fd87-11ea-81de-8d6390198848

This change does what you'd expect - if `rowActions` returns `null`, a toolbar will not be displayed at all

### DIFF
--- a/src/components/TableModule/TableModuleRow.test.tsx
+++ b/src/components/TableModule/TableModuleRow.test.tsx
@@ -105,3 +105,24 @@ test('it renders the provided `rowActions`', async () => {
   const actions = await findByTestId('row-action');
   expect(actions).toBeInTheDocument();
 });
+
+test('it does not render row actions when the `rowActions` function returns null', async () => {
+  const { findByRole } = renderWithTheme(
+    <table>
+      <tbody>
+        <TableModuleRow
+          data={{}}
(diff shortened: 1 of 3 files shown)


src/components/TableModule/TableModuleRow.test.tsx
```diff
@@ -105,3 +105,24 @@ test('it renders the provided `rowActions`', async () => {
   const actions = await findByTestId('row-action');
   expect(actions).toBeInTheDocument();
 });
+
```
HaleyWardo, line 108: Thank you for adding this test! 🙏🏻

HaleyWardo (approved):

mschroering (approved):

PR #19, opened 2020-10-07
CHROM add flex-start and flex-end

HaleyWardo:
**Changes**
- Added `flex-end` and `flex-start`
    - Right now we are setting `flex-end` to `align-items: end` when I think we should allow for both being an option to pass in. 

Ex issue: 

- Before when I passed in `align="flex-end"`
Screen Shot 2020-10-07 at 4 10 09 PM

- After when I passed in `align="flex-end"`
Screen Shot 2020-10-07 at 4 10 14 PM


travomic (approved):

ynotdraw (approved):

ynotdraw:
Nice addition

PR #24, opened 2020-10-16
Add disableDismissOnClickOutside prop to Modal

smolster:
## Motivation
See [this Slack thread](https://lifeomic.slack.com/archives/CHMFL4PQT/p1602771468011800).

## Changes
- Added the `disableDismissOnClickOutside` prop to the `Modal` component, to disable automatic dismissal on clicking the overlay.
- Updated tests + docs around the new prop.
- Added a hard-coded test ID for the overlay component to support unit tests.
  - I couldn't think of an idiomatic way to find that component / execute the click without this. If there's a better alternative to exporting that ID, 100% open to suggestions!


ynotdraw (approved):
This is great! Thanks a bunch for tests and the great documentation additions as well! 🚀

PR #28, opened 2020-10-28
Export useStyles hook

ynotdraw:
### Changes

- Export `useStyles` for each component not already doing so

### Motivation

This would allow folks who need to build a custom component to use the existing styles of a Chroma component.  This would mean the component would _look_ pretty close (if not identical) to a Chroma component, but would serve the developer's custom-built needs.  I believe @HaleyWardo ran into a case where this would've helped her.  She may have more info and a better example

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,3 +91,16 @@ function App({ children }) {
   return <ThemeProvider theme={theme}>{children}</ThemeProvider>;
 }
 ```
+
+## Importing Component Styles Only?
+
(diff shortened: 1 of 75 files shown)


HaleyWardo (approved):
🙌🏻  thank you!

macnaughton (approved):

danperkins (approved):

justin-ligman-lifeomic (approved):

mschroering (approved):

PR #31, opened 2020-10-29
FIX: Allow left align override on last column header and cell

dexterca:
This PR allows for last column table content to be left-aligned. Last column content will continue to default to right-aligned.

BEFORE | AFTER
-- | --
Screen Shot 2020-10-29 at 11 26 19 AM | Screen Shot 2020-10-29 at 11 26 54 AM

### DIFF
--- a/src/components/TableModule/TableModuleCell.test.tsx
+++ b/src/components/TableModule/TableModuleCell.test.tsx
@@ -86,7 +86,7 @@ test('it renders with "isLastCellInRow"', async () => {
     </RenderInTable>
   );
   const root = await findByTestId(testIds.bodyCell);
-  expect(root).toHaveClass('ChromaTableModule-tableRowCellAlignRight');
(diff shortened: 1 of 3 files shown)


HaleyWardo (approved):

src/components/TableModule/TableModuleCell.test.tsx
```diff
@@ -86,7 +86,7 @@ test('it renders with "isLastCellInRow"', async () => {
     </RenderInTable>
   );
   const root = await findByTestId(testIds.bodyCell);
-  expect(root).toHaveClass('ChromaTableModule-tableRowCellAlignRight');
+  expect(root).toHaveClass('ChromaTableModule-tableRowCell');
```
ynotdraw, line 89: Ah interesting.  Yeah this test should've caught this; however, it wasn't written correctly 🙈 .  My b.

ynotdraw (approved):

mschroering (approved):

PR #34, opened 2020-11-02
Add overflow hidden to ensure content doesn't go outside rounded corners

xdumaine:
Gif shows the issue. When a parent has border radius, it needs overflow hidden to ensure child content doesn't exceed the corners.

modal


ynotdraw (approved):
🙇 ty!

PR #38, opened 2020-11-05
DOCS: add data-icon attribute to all icons

ynotdraw:
### Changes

- Add `data-icon` attribute for all icons for documentation folks

I wrote a script that automated all of this, but it seems to have worked fine.  One thing I did notice is that the `write-svgs` script is quite a bit slower.

@jhbeard you'll have to open the dev tools and find the `data-icon` attribute as shown here.  Does this seem helpful enough?

Screen Shot 2020-11-05 at 9 53 26 AM


jhbeard:
That seems perfect. Thanks

jhbeard:
Great solution

jhbeard (approved):

hemp (approved):

PR #43, opened 2020-12-02
Allow for icon being optional in Step defaulting to number position

jkdowdle:
image

I think this could come in handy during prototyping and playing around, not having to worry about providing Icons, and also maybe in some places an Icon just doesn't make sense though mixing the two options probably would not work great

### DIFF
--- a/src/components/Stepper/Step.tsx
+++ b/src/components/Stepper/Step.tsx
@@ -122,7 +122,7 @@ export interface StepProps {
   className?: string;
   completed?: boolean;
   disabled?: boolean;
-  icon: React.ComponentType<React.SVGProps<SVGSVGElement>>;
+  icon?: React.ComponentType<React.SVGProps<SVGSVGElement>>;
   index?: number;
   onClick?: (index: number | undefined) => void;
   subTitle?: string;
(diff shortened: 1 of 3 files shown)


jkdowdle:
@HaleyWardo let me know if there is an accessibility issue to take into consideration, I wasn't quite sure

src/components/Stepper/Step.tsx
```diff
@@ -161,15 +161,29 @@ export const Step: React.FC<StepProps> = ({
...
+            className={clsx(
+              classes.icon,
+              active && classes.activeIcon,
+              completed && classes.completedIcon
+            )}
+            role="img"
+          />
+        ) : (
+          <Text
+            aria-hidden="true"
```
HaleyWardo, line 176: @ynotdraw should we have `aria-hidden` here? Since there is no title or subtitle text for the step, shouldn't we want to allow screen readers to pick up this text?

stories/components/Stepper/default.md
```diff
@@ -85,6 +85,16 @@ component.
...
+### Rendering Steps Ordered as 1, 2, 3, etc
+
+By not providing the Icon, `Step` will display the `index + 1` in the icons place.
+
+```jsx
+<Step title="First" />
+<Step title="Second" />
+<Step title="Last" />
+```
+
```
ynotdraw, line 97: Thanks for updating the docs!

src/components/Stepper/Step.tsx
ynotdraw: Hm to be honest, I'm not quite sure.  I'd have to do some reading on how steppers are expected to work when it comes to a11y. It's _maybe_ okay as-is?

I'll file a task to revisit a11y for this component.

HaleyWardo: Sweet, thanks!

HaleyWardo (approved):

dexterca (approved):

src/components/Stepper/Step.tsx
jkdowdle: @HaleyWardo @ynotdraw Thank you both for taking a look!

PR #46, opened 2020-12-04
a11y: Enforce aria-label if no label/title

ynotdraw:
### Changes

- Enforce `aria-label` as a prop if a `label` or `title` is not provided
- Apply a class that hides the label/title elements from the user, but still accessible via screenreaders

**This is a breaking change**, but is important for accessibility. I pitched this in the Chroma channel and sounded like it was well received 👍 - (let me know if I left someone out on the PR)

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -14,6 +14,7 @@ import { generateUniqueId } from '../_private/UniqueId';
 import { Text } from '../Text';
 import { motion, useMotionValue, useTransform } from 'framer-motion';
 import 'focus-visible';
+import { screenreaderOnlyStyles } from '../../styles/screenreaderOnly';
 
 export const CheckboxStylesKey = 'ChromaCheckbox';
 
@@ -85,6 +86,9 @@ export const useStyles = makeStyles(
       zIndex: 2,
     },
     box: {},
+    srOnly: {
+      ...screenreaderOnlyStyles,
+    },
   }),
   { name: CheckboxStylesKey }
 );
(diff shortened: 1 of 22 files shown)


src/styles/screenreaderOnly.ts
```diff
@@ -0,0 +1,13 @@
...
+  position: 'absolute',
+  width: '1px',
+  height: '1px',
+  padding: 0,
+  margin: '-1px',
+  overflow: 'hidden',
+  clip: 'rect(0, 0, 0, 0)',
+  whiteSpace: 'nowrap',
+  borderWidth: 0,
+};
```
ynotdraw, line 13: Took this from https://tailwindcss.com/docs/screen-readers#class-reference - works well!

src/components/Checkbox/Checkbox.tsx
```diff
@@ -14,6 +14,7 @@ import { generateUniqueId } from '../_private/UniqueId';
 import { Text } from '../Text';
 import { motion, useMotionValue, useTransform } from 'framer-motion';
 import 'focus-visible';
+import { screenreaderOnlyStyles } from '../../styles/screenreaderOnly';
```
ynotdraw, line 17: You'll see below where this is coming from, but is another case where it'd be nice to add things (even full classes) to the theme (https://github.com/lifeomic/chroma-react/issues/47)

jkdowdle (approved):

evanheisler (approved):

mschroering (approved):

Druotic (approved):

RyanRatcliff (approved):

dexterca (approved):

HaleyWardo (approved):

PR #53, opened 2021-01-21
Fix Issue where RadioGroup Did Not Update Value When Prop Changed

sternetj:
Radio Group value was not updating when the value prop changes

### DIFF
--- a/src/components/Radio/RadioGroup.tsx
+++ b/src/components/Radio/RadioGroup.tsx
@@ -99,6 +99,10 @@ export const RadioGroup: React.FC<RadioGroupProps> = ({
 
   const [contextValue, setContextValue] = React.useState(value);
 
+  React.useEffect(() => {
+    setContextValue(value);
+  }, [value]);
+
   const handleContextChange = (e: React.ChangeEvent<HTMLInputElement>) => {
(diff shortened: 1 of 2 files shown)


src/components/Radio/RadioGroup.tsx
```diff
@@ -99,6 +99,10 @@ export const RadioGroup: React.FC<RadioGroupProps> = ({
 
   const [contextValue, setContextValue] = React.useState(value);
 
+  React.useEffect(() => {
+    setContextValue(value);
+  }, [value]);
```
ynotdraw, line 104: Hmm, I'm not sure I'm following why the existing code doesn't cover this?  If a radio option in the radio group is clicked, the `handleContextChange` should be called, calling `setContextValue` as well, right?

sternetj: Yes it is. I have a form that I am working with and once the form is reset the radio group is not respecting setting it back to the default value

ynotdraw: Ah okay, makes sense! Thanks!

ynotdraw (approved):

PR #57, opened 2021-01-26
PHC-463 Select aria-describedby

HaleyWardo:
**Changes**
- Conditionally add `aria-describedBy` attribute if `hasError` or `helpMessage`.

Screen Shot 2021-01-26 at 2 59 16 PM

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -12,7 +12,6 @@ import { makeStyles } from '../../styles';
 import { GetClasses } from '../../typeUtils';
 import {
   BaseFormElement,
-  buildDescribedBy,
   errorFor,
   FormErrorMessage,
   FormHelpMessage,
@@ -425,7 +424,6 @@ export const Select: React.FC<SelectProps> = ({
           }
         )}
         ref={buttonRef}
-        aria-describedby={buildDescribedBy(uniqueId)}
         id={uniqueId}
         {...popover}
         {...rootProps}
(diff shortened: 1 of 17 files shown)


src/components/Select/Select.tsx
```diff
@@ -425,7 +424,6 @@ export const Select: React.FC<SelectProps> = ({
           }
         )}
         ref={buttonRef}
-        aria-describedby={buildDescribedBy(uniqueId)}
```
ynotdraw, line 428: Actually, I believe this is used for the help/error messages.  Let's see if we can conditionally apply this attribute, if/when we do have a help or error message?

ynotdraw (changes requested):
Will pair on changes mentioned in comment! I think conditionally applying this should work. We may also need to do it for other form elements. Nice find!

ynotdraw:
FWIW, just verified with Haley that these changes work with a screenreader 👍

src/components/TextField/TextField.test.tsx
```diff
@@ -19,7 +19,7 @@ test('it renders a TextField', async () => {
   const textfield = await findByTestId(testId);
   expect(textfield).toBeInTheDocument();
   expect(textfield).toHaveClass('ChromaTextField-input');
-  expect(textfield.getAttribute('aria-describedby')).toBeTruthy();
+  expect(textfield.getAttribute('aria-describedby')).toBeFalsy();
```
ynotdraw, line 22: These are a bit redundant as we have the other tests, but that's okay.

ynotdraw (approved):
🙇

dexterca (approved):

simons5593 (approved):

PR #60, opened 2021-01-28
PHC-463 TextField label

HaleyWardo:
**Changes**
- Conditionally rendered label and added `aria-label` to input field

Screen Shot 2021-01-28 at 2 47 58 PM

### DIFF
--- a/src/components/TextField/TextField.test.tsx
+++ b/src/components/TextField/TextField.test.tsx
@@ -227,6 +227,24 @@ test('it renders an inverse color icon when icon and tooltipMessage are provided
   expect(icon).toHaveClass('ChromaTextField-labelIconInverse');
 });
 
+test('it renders label when provided', async () => {
+  const { findByLabelText } = renderWithTheme(
+    <TextField label="label-text" data-testid={testId} />
+  );
+
+  const label = await findByLabelText(/label-text/);
+  expect(label).toBeInTheDocument();
+});
+
+test('it does not render label when not provided', async () => {
+  const { queryByLabelText } = renderWithTheme(
+    <TextField label="" data-testid={testId} />
+  );
+
+  const label = await queryByLabelText(/label-text/);
+  expect(label).not.toBeInTheDocument();
+});
+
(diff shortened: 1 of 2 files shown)


dexterca (approved):

src/components/TextField/TextField.test.tsx
```diff
@@ -227,6 +227,24 @@ test('it renders an inverse color icon when icon and tooltipMessage are provided
...
+  expect(label).toBeInTheDocument();
+});
+
+test('it does not render label when not provided', async () => {
+  const { queryByLabelText } = renderWithTheme(
+    <TextField label="" data-testid={testId} />
+  );
+
+  const label = await queryByLabelText(/label-text/);
+  expect(label).not.toBeInTheDocument();
```
ynotdraw, line 245: Something else that may work, if you want to ensure no `label` elements at all is:

```suggestion
renderWithTheme(
  <TextField label="" data-testid={testId} />
);

expect(document.querySelectorAll('label').length).toEqual(0);
```

src/components/TextField/TextField.tsx
```diff
@@ -200,42 +196,44 @@ export const TextField = React.forwardRef<HTMLInputElement, TextFieldProps>(
...
-              </span>
-            </Tooltip>
-          )}
-          {secondaryLabel && (
-            <span className={classes.labelSecondary}>{secondaryLabel}</span>
-          )}
-        </label>
+        {label && (
+          <label
+            aria-hidden="true"
```
ynotdraw, line 201: I'm curious if we can/should delete this as well?

simons5593 (approved):

src/components/TextField/TextField.test.tsx
HaleyWardo: Oh sweet! I will update this.

src/components/TextField/TextField.tsx
HaleyWardo: I don't think we need it anymore.

ynotdraw (approved):

PR #61, opened 2021-01-29
PHC-463 Button Link hover

HaleyWardo:
Issue:  Bootstrap styles override button color set in root styles when hovering on `ButtonLink`. 

Screen Shot 2021-01-29 at 8 47 33 AM

**Changes**
- Added `color: white` when hovering over button

**BEFORE**
Screen Shot 2021-01-29 at 8 38 21 AM

**AFTER**
Screen Shot 2021-01-29 at 8 44 03 AM

### DIFF
--- a/src/components/ButtonLink/ButtonLink.tsx
+++ b/src/components/ButtonLink/ButtonLink.tsx
@@ -41,6 +41,7 @@ export const useStyles = makeStyles(
       userSelect: 'none',
       '&:hover, &:focus': {
         textDecoration: 'none',
+        color: theme.palette.common.white,
         backgroundColor: theme.palette.primary[800],
         outline: 'none',
       },


HaleyWardo:
I wasn't sure if this should just be handled in the app or if this was okay to update in Chroma. @ynotdraw do you think this is okay?

dexterca:
This should be good to update in Chroma. It's been on my list for awhile. Thanks for picking it up.

simons5593 (approved):

dexterca (approved):

ynotdraw (approved):
Yeah, I think this is specific to LifeOmic apps as we pull in a CSS file that styles on top of base elements.  As you can see here, it doesn't do this when you _only_ use Chroma; however, adding this in here certainly doesn't hurt anything, as the CSS file is overriding the base _hover_ `color` of the root element  👎. 

https://user-images.githubusercontent.com/8069555/106284775-9f35c780-6211-11eb-9f02-822ca6b0ff28.mov


**TLDR; makes sense to have this in, as it's in the base class, but anything overriding `<a` tag hover styles will override our defaults.**. Thank you! 🙇

PR #68, opened 2021-03-03
Set text color of popover to primary to complement background color

xdumaine:
Fixes a bug where if popover is a child of a div with `color: white` (or other non-contrasting text with it's white background), the text is not visible. Since it sets the background explicitly to white, it should also set the text color.

Bug in action:

white-text


ynotdraw (approved):
Nice find, thanks for the fix!

PR #76, opened 2021-03-11
WELL-706 - Support for grouped select options

jagoda:


src/components/Select/GroupHeading.tsx
```diff
@@ -0,0 +1,36 @@
+import clsx from 'clsx';
+import * as React from 'react';
+import { makeStyles } from '../../styles';
+import { GetClasses } from '../../typeUtils';
+
+export const GroupHeadingStylesKey = 'ChromaSelectGroupHeading';
+
+export const useStyles = makeStyles(
+  (theme) => ({
+    root: {
```
jagoda, line 10: Styling is not exactly my strong suit... I tried to follow some examples in the code but open to any feedback on how to make this better.

src/components/Select/Select.tsx
```diff
@@ -370,6 +371,10 @@ export const Select: React.FC<SelectProps> = ({
         return null;
       }
 
+      if (child.type === GroupHeading) {
```
jagoda, line 374: I was able to find the `type` attribute in dev forum conversations and looked at the React source behind it. This seems to do what I want, but I didn't see it called out explicitly anywhere in the React docs. Not sure how we feel about that :confused:

src/components/Select/GroupHeading.tsx
ynotdraw: Sure thing! I'll take a look this evening and can provide any feedback. We may defer to Casey as well for additional polish post-merge!

ynotdraw: Just chatted with @dexterca, sounds like he may have some time tomorrow to style this out a bit!

src/components/Select/Select.tsx
ynotdraw: I think this is okay for now after searching around a bit as well.  I see a few other libraries doing this too.  If React ends up pulling this at some point, hopefully they'll provide a new, documented replacement! 👍

ynotdraw (changes requested):
Last bit and then this looks 💯 - if the overrides can be output from https://github.com/lifeomic/chroma-react/blob/17698becef207877481de72b325d75a0ffda86aa/src/styles/overrides/ChromaOverrides.ts like the other components, then you're set!

jagoda:
Oops, thanks. Done!

ynotdraw (approved):
Thank you! 🚀

PR #77, opened 2021-03-12
WELL-706 - Fix GroupHeading rendering

jagoda:
The built code loses the `type` attribute information. Using a
data-* attribute is preserved across the build.

### DIFF
--- a/src/components/Select/GroupHeading.tsx
+++ b/src/components/Select/GroupHeading.tsx
@@ -28,11 +28,13 @@ export type GroupHeadingClasses = GetClasses<typeof useStyles>;
 
 export interface GroupHeadingProps {
   className?: string;
+  ['data-select-role']: 'heading';
 }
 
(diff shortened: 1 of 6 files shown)


src/components/Select/GroupHeading.tsx
```diff
@@ -28,11 +28,13 @@ export type GroupHeadingClasses = GetClasses<typeof useStyles>;
 
 export interface GroupHeadingProps {
   className?: string;
+  ['data-select-role']: 'heading';
 }
 
 export const GroupHeading: React.FC<GroupHeadingProps> = ({
   children,
   className,
+  ['data-select-role']: dataSelectRole,
```
jagoda, line 37: I did this just to keep the select role from being forwarded to the `li` element.

ynotdraw: Works for me 👍

ynotdraw (approved):

PR #90, opened 2021-03-17
Update secondary (green) and error (red) palettes

dexterca:
Both the green and red palettes had some anomalies in them to where as you progressed through them the values didn't follow a logical pattern. I believe this was the case because we forced the red and green from PHC's logo into their respective palettes.

With is PR, the red and green palettes are independent of the PHC logo palette and now follow a logical pattern.

### BEFORE ➡️ AFTER

Screen Shot 2021-03-16 at 7 56 45 PM


ynotdraw (approved):
🎨 🔥

edoctonLO:
Looks awesome!

PR #103, opened 2021-03-18
Get TextArea story to pass a11y checks

dexterca:
Add aria labels to TextArea demo so a11y checks pass

### DIFF
--- a/stories/components/TextArea/TextArea.stories.tsx
+++ b/stories/components/TextArea/TextArea.stories.tsx
@@ -29,7 +29,7 @@ const TextAreaStory: React.FC = () => {
               setText(e.target.value);
             }}
             label="Name"
-            />
+          />
           <TextArea label="Name" secondaryLabel="Optional" />
           <TextArea
             aria-label="Name"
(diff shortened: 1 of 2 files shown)


stories/components/TextArea/TextArea.stories.tsx
```diff
@@ -31,33 +32,38 @@ const TextAreaStory: React.FC = () => {
           />
           <TextArea label="Name" secondaryLabel="Optional" />
           <TextArea
+            aria-label="Name"
             label="Name"
```
ynotdraw, line 36: Should `aria-label` only be required on the text areas without a `label`? Or does it complain?

dexterca: It's alerting this on every `<TextArea` Any ideas?

Screen Shot 2021-03-18 at 1 13 19 PM

HaleyWardo: I am wrapping up a PR and then I can take a quick look at why this is happening.

ynotdraw: I think it may actually be `aria-hidden="true"` - I think we can get rid of that if I'm not mistaken 

https://github.com/lifeomic/chroma-react/blob/feature/color-contrast/src/components/TextArea/TextArea.tsx#L201

dexterca: 👀

dexterca: That did the trick, thanks @ynotdraw

ynotdraw: Any time! Not sure why I added that to begin with, so sorry for the troubles

ynotdraw (approved):

PR #112, opened 2021-04-07
WELL-933 - Don't hide modal actions

jagoda:
With the addition of `overflow: 'hidden'` the modal actions can
be clipped on small screens.

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -46,6 +46,8 @@ export const useStyles = makeStyles(
     content: {
       background: theme.palette.common.white,
       borderRadius: theme.pxToRem(10),
+      display: 'flex',
+      flexDirection: 'column',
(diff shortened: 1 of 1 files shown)


jagoda:
Before:
image

After:
image

For a "real world" example see lifeomic/wellness-admin-app#209.

ynotdraw (approved):
Ah, interesting. Thanks!  I think this is good as-is, but maybe something @HaleyWardo and I could look at later:

Should we do a `max-height: <some-reasonable-value>vh` or something on the content element [here](https://github.com/lifeomic/chroma-react/blob/master/src/components/Modal/Modal.tsx#L348) so that we ensure from within Chroma that the height of the content-area doesn't get too tall?  Maybe this could be done on mobile-only or lower resolutions?  Or maybe it makes sense to do across the board?

PR #128, opened 2021-05-18
PHC-1157

mattalco:
Correcting a minor icon misalignment on IE.

2px comes from the 2rem height of the container minus the 6px of padding (for a total height of 26px).


ynotdraw:
Looks good to me! Thanks for the contribution and IE-specific fix!

ynotdraw (approved):

PR #129, opened 2021-05-19
PHC-1172: updating styles for Select and Combobox (IE 11)

mattalco:
closes PHC-1172


ynotdraw (approved):
Nice! Thanks for the IE11 fixes!

PR #141, opened 2021-06-11
Add number format inputs from connect-web-console

zhengkyl:
## Changes
This adds the number input fields from https://github.com/lifeomic/connect-web-console/blob/master/src/elements/Form/NumberFormats.test.tsx. All the tests are included as well, with minor changes due to the different testing library.

## Purpose
These input fields are useful for multiple projects, including `wellness-admin-app`.

## What it looks like
image

### DIFF
--- a/src/components/NumberFormat/UnitNumberFormatField.tsx
+++ b/src/components/NumberFormat/UnitNumberFormatField.tsx
@@ -3,13 +3,21 @@ import { TextFieldProps, TextField } from '../TextField';
 
 import { makeStyles } from '../../styles';
 import clsx from 'clsx';
+import { GetClasses } from '../../typeUtils';
 
-const useStyles = makeStyles((theme) => ({
-  textField: {
-    marginTop: theme.spacing(2),
-    marginBottom: theme.spacing(1),
-  },
-}));
+export const UnitNumberFormatFieldStylesKey = 'ChromaUnitNumberFormatField';
+
+const useStyles = makeStyles(
+  (theme) => ({
+    textField: {
+      marginTop: theme.spacing(2),
+      marginBottom: theme.spacing(1),
+    },
+  }),
+  { name: UnitNumberFormatFieldStylesKey }
+);
+
+export type UnitNumberFormatFieldClasses = GetClasses<typeof useStyles>;
 
 export const UnitNumberFormatField: React.FC<
   Omit<TextFieldProps, 'onChange' | 'value'> & {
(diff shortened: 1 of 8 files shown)


src/components/NumberFormat/UnitNumberFormatField.tsx
```diff
@@ -0,0 +1,198 @@
...
+    prefixUnits = false,
+    className,
+    ...other
+  } = props;
+  // formatting helper. Prefixes units if specified, otherwise suffix.
+  // also automatically adds a space if the units string is > 2 e.g $10.00 30%, 12ml, 10 slots
+  const format = (val: number) =>
+    prefixUnits
+      ? `${units}${val.toFixed(decimalScale)}`
+      : `${val.toFixed(decimalScale)}${units.length > 2 ? ' ' : ''}${units}`;
```
ynotdraw, line 42: Thinking forward a bit, I'm curious how this will change if we need to support RTL.  Maybe go through a provided `format` prop?  No need to change now, just something we can keep in the back of our minds 👍

ynotdraw (changes requested):
Hey @zhengkyl this is looking really great!  The tests are awesome.  Great work on getting these added! 💯 

To ensure users can override the styles for these components at the global level, the styles key and classes should be exported here: https://github.com/lifeomic/chroma-react/blob/master/src/styles/overrides/ChromaOverrides.ts

We give the components a name via: https://github.com/lifeomic/chroma-react/blob/master/src/components/Alert/Alert.tsx#L7

and provide it to: https://github.com/lifeomic/chroma-react/blob/master/src/components/Alert/Alert.tsx#L38

src/components/NumberFormat/UnitNumberFormatField.tsx
zhengkyl: That's a good observation. There is also the case where currency symbols are not consistently on one side within LTR and RTL languages.

ynotdraw (approved):
Perfect! Thanks a bunch @zhengkyl , great work!

src/components/NumberFormat/UnitNumberFormatField.tsx
ynotdraw: Ah, nice insight on currency symbols as well.  We should support this eventually, but I think this is good to go as is if y'all don't need to support those cases just yet 👍.  We will get a lot of value out of these as is, and whoever needs to support that first (maybe Lifeology?) can add that functionality and the rest of the products benefit!

PR #143, opened 2021-06-11
Deprecate Chroma icons in favor of Chromicons

ynotdraw:
### Changes

- Deprecate the lined + linedCustom icons coming from Chroma in favor of using Chromicons directly.  In a future version, Chroma will also switch to using Chromicons directly internally.

**NOTE:** This should not block you from still using the icons from Chroma, it's more nudging you to use Chromicons directly instead.  We will let this bake a while before ripping these icons out and release with a proper breaking change (minor since we are still pre-1.0).

Here is what the deprecation notice will look like if you use VS Code:

dep-notice

### DIFF
--- a/src/icons/lined/Activity.tsx
+++ b/src/icons/lined/Activity.tsx
@@ -1,6 +1,13 @@
 import { Activity as FeatherActivity, Props } from 'react-feather';
 import * as React from 'react';
 
+/**
+ * @deprecated It is recommended you use the Chromicons package directly (https://github.com/lifeomic/chromicons)
+ * instead of the icons provided by Chroma.
+ *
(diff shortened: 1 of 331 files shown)


dexterca (approved):

macnaughton (approved):
image

smolster (approved):
😮

xdumaine (approved):

namick (approved):

justin-ligman-lifeomic:
Well, I was going to approve this PR as well, but it keeps crashing Chrome every time I switch over to the changes, so no approve for you!

ynotdraw:
Will be adding [chromicons](https://github.com/lifeomic/chromicons) to Chroma after a bit 👍

PR #147, opened 2021-06-14
v0.5.4

zhengkyl:
Release number inputs from #141. (I forgot to use the `version` command in last PR)


smolster (changes requested):
I think you can pull off what you want without an additional tag here. I'll ping you directly.

smolster (approved):
So be it. RIP v0.5.3

PR #149, opened 2021-06-15
[Proposal] Adopt semantic-release

smolster:
**Context**: https://lifeomic.slack.com/archives/CHMFL4PQT/p1623699680032800

## Motivation
I'd love to propose adopting [`semantic-release`](https://github.com/semantic-release/semantic-release#readme) to release this package declaratively. **This would eliminate the need for "version bump PRs", instead enforcing that all PRs have at least 1 commit with a message matching [the conventional commits spec](https://www.conventionalcommits.org/en/v1.0.0/#summary).**

**Prior Art**
We've adopted this strategy in most (if not all) package repos maintained by the SkillSpring team, to great effect. The changeset proposed here is exactly the same as those we've used to adopt this strategy in other packages ([example](https://github.com/lifeomic/connect-common-ui/pull/78)).

**How it Works**
There are two main pieces here:
- `release.config.js`, the local config for `semantic-release`. This config directs semantic-release to:
  - use the conventional commits spec
  -  publish a new npm version and GitHub release when it determines a new release should be made. It also comments on merged PRs, specifying which version the PR was released in ([example](https://github.com/lifeomic/connect-common-ui/pull/117#issuecomment-840556335)).
- `lifeomic-probot.yml`, [a `lifeomic-probot` configuration](https://github.com/lifeomic/probot#enforcesemanticcommits) that checks PRs to ensure they have at least one conventional commit. It leaves helpful messages when they do not, and also flags PRs with breaking changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,9 @@
   "repository": "git@github.com:lifeomic/chroma-react.git",
   "author": "LifeOmic <oss@lifeomic.com>",
   "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
   "scripts": {
     "build": "run-p build:*",
     "build:esm": "tsc -p tsconfig.build.json -m es2015 --outDir ./dist/esm && cp -r ./src/assets/ ./dist/esm/assets",
@@ -56,6 +59,7 @@
     "@typescript-eslint/eslint-plugin": "^4.0.1",
     "@typescript-eslint/parser": "^4.0.1",
     "babel-loader": "^8.1.0",
+    "conventional-changelog-conventionalcommits": "^4.6.0",
     "eslint": "^7.8.0",
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-jsx-a11y": "^6.3.1",
@@ -74,6 +78,7 @@
     "react-router-dom": "^5.1.2",
     "react-test-renderer": "^16.13.1",
     "reakit-test-utils": "^0.11.1",
+    "semantic-release": "^17.4.4",
     "storybook-readme": "^5.0.8",
     "tap-xunit": "^2.4.1",
     "ts-jest": "^25.4.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@lifeomic/chroma-react",
-  "version": "0.5.4",
+  "version": "0.0.0",
   "description": "Chroma Design System",
   "sideEffects": false,
   "repository": "git@github.com:lifeomic/chroma-react.git",
--- a/package.json
+++ b/package.json
@@ -31,12 +31,14 @@
     }
   },
   "peerDependencies": {
+    "@lifeomic/chromicons": "^0.2.1",
     "react": "16",
     "react-dom": "16",
     "react-router-dom": ">=5.1.2"
   },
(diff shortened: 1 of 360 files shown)


package.json
```diff
@@ -6,6 +6,9 @@
   "repository": "git@github.com:lifeomic/chroma-react.git",
   "author": "LifeOmic <oss@lifeomic.com>",
   "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
```
smolster, line 11: This is the solution semantic-release recommends for handling the `--access public` arg we were passing to `npm publish` before.

[docs](https://semantic-release.gitbook.io/semantic-release/support/faq#how-can-i-set-the-access-level-of-the-published-npm-package)

.github/workflows/pr-branch-build.yml
```diff
@@ -19,3 +19,10 @@ jobs:
           yarn lint
           yarn test
           yarn build
+      - name: Dry Run Publish
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{secrets.LIFEOMIC_NPM_TOKEN}}
+        run: |
+          yarn build
+          npx semantic-release --dry-run
```
smolster, line 28: Using `--dry-run` here will cause it to log out the "expected" version bump. Pretty nifty.

package.json
```diff
@@ -1,11 +1,14 @@
 {
   "name": "@lifeomic/chroma-react",
-  "version": "0.5.4",
+  "version": "0.0.0",
```
smolster, line 3: Since this `version` entry is no longer a source of truth, we've taken the approach of setting it to `0.0.0` for clarity.

xdumaine (dismissed):
This'll be great for automatic versioning. Only question - does it handle pre-1.0.0 releases correctly? I.e., `breaking` does a minor change?

smolster:
> does it handle pre-1.0.0 releases correctly

Great question, just did some testing. It seems like the answer is no, so we'll want to use `feat` for breaking changes until we this project gets to 1.0.0.

ynotdraw (approved):
Thanks for doing this!  This is neat.  One thing that may be helpful before merging is letting the `#chroma` channel in Slack know as a heads up about the new process for commit messages + releases so there isn't any confusion.

I suppose there could also be a note either in the `README.md` or via a new `CONTRIBUTING.md`?

smolster:
> I suppose there could also be a note/documentation either in the README.md or via a new CONTRIBUTING.md?

Definitely. I've PR'd a runbook in the wiki. Once that merges, I'll update the documentation here to point there. (https://github.com/lifeomic/wiki/pull/235)

smolster:
Actually, I'll add more explicit documentation in this repo, and avoid linking to the wiki page, since this repo is public and the wiki is not.

ynotdraw:
> Actually, I'll add more explicit documentation in this repo, and avoid linking to the wiki page, since this repo is public and the wiki is not.

Awesome! I like your thinking! Thanks!

xdumaine (approved):

dexterca (approved):

HaleyWardo (approved):

PR #151, opened 2021-06-22
Color picker

HaleyWardo:
**Changes**
- Added color picker
 
https://user-images.githubusercontent.com/32574227/122842556-a007ec80-d2cb-11eb-8b85-441765839f12.mov

### DIFF
--- a/src/components/ColorPicker/ColorPicker.tsx
+++ b/src/components/ColorPicker/ColorPicker.tsx
@@ -1,0 +1,461 @@
+import clsx from 'clsx';
+import * as React from 'react';
+import { makeStyles, useTheme } from '../../styles';
+import { GetClasses } from '../../typeUtils';
+import {
+  BaseFormElement,
+  buildDescribedBy,
+  errorFor,
+  FormErrorMessage,
+  FormHelpMessage,
+  helpFor,
+} from '../_private/forms';
+import { generateUniqueId } from '../_private/UniqueId';
+import { Tooltip } from '../Tooltip';
+import { Popover, PopoverItem, PopoverList } from '../Popover';
+import { ButtonUnstyled } from '../ButtonUnstyled';
+
+export const ColorPickerStylesKey = 'ChromaColorPicker';
+
+export const useStyles = makeStyles(
+  (theme) => ({
+    root: {},
+    label: {
+      alignItems: 'center',
+      color: theme.palette.black[800],
+      display: 'flex',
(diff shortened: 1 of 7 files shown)


HaleyWardo:
I paused on adding this because I was running into some performance issues. Instead, I decided to put this component into PHC. After adding it in this morning, @ynotdraw pointed out that maybe the performance issue was caused because the story was rendering a lot of `ColorPicker`s... which was the case. 🤦‍♀️  Once this is approved I will update this [PR](https://github.com/lifeomic/phc-ui/pull/1229) to just use this component.

benkraus (approved):

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,470 @@
...
+              <ButtonUnstyled
+                aria-label="Pick color"
+                disabled={disabled || readOnly}
+              >
+                <Color className={classes.colorPosition} color={colorValue} />
+              </ButtonUnstyled>
+            }
+            aria-label="Color Picker"
+          >
+            {({ popover }: { popover: any }) => (
```
ynotdraw, line 406: I could be wrong, but I think we export a type for the render props here:

```suggestion
            {({ popover }: PopoverRenderProps) => (
```

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,470 @@
...
+  errorFor,
+  FormErrorMessage,
+  FormHelpMessage,
+  helpFor,
+} from '../_private/forms';
+import { ButtonUnstyled } from '../ButtonUnstyled';
+import { generateUniqueId } from '../_private/UniqueId';
+import { GetClasses } from '../../typeUtils';
+import { makeStyles, useTheme } from '../../styles';
+import { Popover, PopoverItem, PopoverList } from '../Popover';
```
ynotdraw, line 15: In combination with below:

```suggestion
import { Popover, PopoverItem, PopoverList, PopoverRenderProps } from '../Popover';
```

src/components/ColorPicker/ColorPicker.tsx
HaleyWardo: Ooo for some reason I only tried `PopoverReturnProps`. Thank you! 🙏

xdumaine (approved):

ynotdraw (approved):

PR #155, opened 2021-06-22
Convert PriceFormatField to use penny values

zhengkyl:
This changes the `PriceFormatField` props of `min`, `max`, `value`, `onChange` to deal with values in terms of integer penny amounts instead of decimal dollar amounts. The underlying `UnitNumberFormatField` still uses decimal numbers because it should be able to handle anything, but this should be hidden away when dealing with money that only ever deals with integer penny amounts.

This is motivated by a desire to standardize dealing with money using pennies, to remove confusion and the need for converting when interfacing with API's that expect penny amounts. This is technically a breaking change, but since I don't believe anyone else is using this yet, hopefully making this change now will prevent confusion in the future.

### DIFF
--- a/src/components/NumberFormat/UnitNumberFormatField.test.tsx
+++ b/src/components/NumberFormat/UnitNumberFormatField.test.tsx
@@ -351,11 +351,50 @@ test('rendering a PercentFormatField renders a UnitNumberFormatField with % unit
 
 test('rendering a PriceFormatField renders a UnitNumberFormatField with $ prefixed units', async () => {
   const onChange = jest.fn();
-  const value = 10;
+  const value = 1000;
   const view = renderWithTheme(
     <PriceFormatField value={value} onChange={onChange} />
   );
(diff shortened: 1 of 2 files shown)


xdumaine (approved):
Great! Agree about technically breaking, but I'm ok with it since your in progress work is the only consumer so far (plus chroma's still pre-1.0.0). One extra way to really vet these changes is to adopt this in connect-web-console (if you look at the output of the jenkins build from your PR you should get an npm prepublish version at the very end which you can `yarn add` to connect-web console to try it out.

smolster (commented):
Changes look great to me!

~I think we should record this change somewhere so consumers know exactly what changes they need to make when upgrading. Maybe @ynotdraw has an existing approach for this? e.g. we use [`UPGRADING.md`](https://github.com/lifeomic/typescript-tools/blob/master/UPGRADING.md) in other projects~

nvm, forgot that these components aren't used anywhere 😄

smolster (approved):

PR #159, opened 2021-06-24
FEAT: Add chromicons and use them over internal icons

ynotdraw:
### Changes

- Use Chromicons internally instead of the now deprecated, internal icons (https://github.com/lifeomic/chroma-react/pull/143)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,12 +28,14 @@
     }
   },
   "peerDependencies": {
+    "@lifeomic/chromicons": "^0.2.1",
     "react": "16",
     "react-dom": "16",
     "react-router-dom": ">=5.1.2"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0-0",
+    "@lifeomic/chromicons": "^0.2.1",
     "@material-ui/types": "^5.0.1",
     "@storybook/addon-a11y": "^5.3.18",
     "@storybook/addon-actions": "^5.3.18",
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -1,6 +1,6 @@
 import clsx from 'clsx';
 import * as React from 'react';
-import { X } from '../../icons/lined';
+import { X } from '@lifeomic/chromicons';
 import { makeStyles } from '../../styles';
 import { GetClasses } from '../../typeUtils';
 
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -82,6 +82,7 @@
     "typescript": "^3.8.3"
   },
   "dependencies": {
+    "@lifeomic/chromicons": "^0.2.1",
```
smolster, line 85: Asking, just for completeness. Should this be a `peerDependency`? I suppose the only concrete benefit would be that consumers could get new icons released in new minor versions of `chromicons`.

smolster (approved):

dexterca (approved):

package.json
ynotdraw: I _almost_ put it as a `peerDependency`, but then decided against it for no good reason other than my initial gut feeling 😂 .  I'm happy to add it as a `peerDependency` though!  What are your thoughts here?  Think it'd be better as a `peerDependency`?

smolster: I suppose it _shouldn't_ matter, since it's just components. multiple copies wouldn't be a problem, and _should_ resolve at runtime in an expected manner, right? (i.e. `chroma` components would get their dependent version of chromicons, and consumer imports should get their (potentially different) version)

ynotdraw: Exactly my thinking and why I hesitated with `peerDependency`.  Chroma doesn't use a ton of icons internally, but the apps themselves do, so if there is a mismatch in the versioning of Chromicons, it shouldn't be a big deal if Chroma is "behind".

smolster: If we change our mind in the future, it will be easier to migrate _away_ from peer dependency than _to_ peer dependency. (_to_ peer dependency is a breaking change for consumers)

ynotdraw: > If we change our mind in the future, it will be easier to migrate _away_ from peer dependency than _to_ peer dependency. (_to_ peer dependency is a breaking change for consumers)

Love it.  Thanks for hashing this out with me 🙇 .  PR incoming!

zhengkyl (approved):

smolster (approved):

macnaughton (approved):

HaleyWardo (approved):

dexterca (approved):

PR #161, opened 2021-06-25
Adjust margins so single line of chips does not affect ComboBox height

dexterca:
When ComboBox had a single line of chips, the height of the parent wrapper would increase by 1px once a chip(s) would appear.

**BEFORE WITHOUT CHIPS** | **BEFORE WITH CHIPS** | **AFTER WITHOUT CHIPS** | **AFTER WITH CHIPS**
-- | -- | -- | --
Screen Shot 2021-06-25 at 11 43 55 AM | Screen Shot 2021-06-25 at 11 46 23 AM | Screen Shot 2021-06-25 at 11 51 12 AM | Screen Shot 2021-06-25 at 11 51 27 AM
`height: 35px` (computed) | `height: 36px` (computed) | `height: 35px` (computed) | `height: 35px` (computed)

One side effect of this change is that chips are no longer perfectly centered vertically. In other words there is one pixel more of padding at the bottom of the chip list than at the top of the chip list (as seen below). So the trade off here is the height of the `<Select` will no longer shift in height once a chip appears but the chips are no longer vertically centered. 😢 

Screen Shot 2021-06-25 at 11 54 57 AM

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -205,7 +205,7 @@ export const useStyles = makeStyles(
     chipList: {
       display: 'flex',
       flexFlow: 'wrap',
-      margin: theme.spacing(0, -0.5),
+      margin: theme.spacing(-0.5, -0.5, 0),
     },
     chip: {
       alignItems: 'center',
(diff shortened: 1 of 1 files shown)


ynotdraw (approved):
RIP, makes sense to me though! Thanks!

smolster (approved):

HaleyWardo (approved):

PR #166, opened 2021-07-06
FIX: add backgroundImage prop to PageLayout

HaleyWardo:
**Changes**
- Added `backgroundImage` prop to PageLayout
  - It can be overridden at a theme level but we should also allow one time overrides

### DIFF
--- a/stories/components/PageLayout/applyBackground.md
+++ b/stories/components/PageLayout/applyBackground.md
@@ -21,7 +21,7 @@ import { PageLayout } from '@lifeomic/chroma-react/components/PageLayout';
 
 ### Overriding The Image
 
-To override the image, leverage the styles key:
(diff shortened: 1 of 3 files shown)


stories/components/PageLayout/applyBackground.md
```diff
@@ -32,6 +32,18 @@ ChromaPageLayout: {
...
+
+```jsx
+<PageLayout
+  title="Account 1"
+  applyBackgroundCover
+  backgroundImage={overrideImage}
+>
+  {/* Your content */}
+</PageLayout>
+```
```
ynotdraw, line 45: Cool thanks for updating the docs!

ynotdraw (approved):

smolster (approved):

PR #167, opened 2021-07-07
fix: increase Select maxHeight to 328px

dexterca:
Some consumers have concerns that the `maxHeight` of the `<Select` popover is too small. There are probably a couple different ways to solve this but this seemed the most simplistic since most mobile devices are taller than 328px.

If anyone has any other ideas I'm open to other options.

The complaint we are getting is that with rating-type questions that require ten or more options, the user might think the ratings stop at four when in fact options go up to ten. 

BEFORE | AFTER
-- | --
Screen Shot 2021-07-07 at 2 52 37 PM | Screen Shot 2021-07-07 at 2 18 03 PM

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -184,7 +184,7 @@ export const useStyles = makeStyles(
       display: 'block',
       listStyle: 'none',
       margin: 0,
-      maxHeight: theme.pxToRem(168),
+      maxHeight: theme.pxToRem(328),
       overflowY: 'auto',
       padding: theme.spacing(1, 0),
     },


xdumaine:
Wow in that before it's hard to tell that it scrolls at all.

smolster (approved):
If there's an easy way to just briefly show the scroll bar on opening, that might be enough communication. But, this fix is definitely fine with me.

HaleyWardo:
> If there's an easy way to just briefly show the scroll bar on opening, that might be enough communication. But, this fix is definitely fine with me.

Yeah when Mel originally showed me this ticket, I asked her to reach out to you (@dexterca) about this one. I told her it might be better to show some sort of indicator that there is more options to display. Not exactly if that is doable/wanted but figured that might be a good option.

dexterca:
> > If there's an easy way to just briefly show the scroll bar on opening, that might be enough communication. But, this fix is definitely fine with me.
> 
> Yeah when Mel originally showed me this ticket, I asked her to reach out to you (@dexterca) about this one. I told her it might be better to show some sort of indicator that there is more options to display. Not exactly if that is doable/wanted but figured that might be a good option.

Yeah I think the best solution (but more complicated) would be to have something similar to Slack's  _⬇️ More unreads_ in the the sidebar. Someone with more JS chops would need to pitch in on this solution though.

ynotdraw:
> > > If there's an easy way to just briefly show the scroll bar on opening, that might be enough communication. But, this fix is definitely fine with me.
> 
> > 
> 
> > Yeah when Mel originally showed me this ticket, I asked her to reach out to you (@dexterca) about this one. I told her it might be better to show some sort of indicator that there is more options to display. Not exactly if that is doable/wanted but figured that might be a good option.
> 
> 
> 
> Yeah I think the best solution (but more complicated) would be to have something similar to Slack's  _⬇️ More unreads_ in the the sidebar. Someone with more JS chops would need to pitch in on this solution though.

An intersection observer may come in handy here.

https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

ynotdraw (approved):
I'm good with this for now if we need a quick fix. On mobile devices, will this be okay as well (can't remember typical sizes off the top of my head)?

PR #168, opened 2021-07-08
Add a default type on makeStyles

xdumaine:
Motivation: Typescript 4 makes the type inference on `props` more strict, and where the type of props cannot be strictly inferred, consumers must add a a type parameter to satisfy makeStyles. Some do `makeStyles<{}>(...)` `makeStyles<Record<string, string>>(...)` or `makeStyles<uknown>()` for simplicity. Adding a default type ([matching the types in mui](https://github.com/mui-org/material-ui/blob/v4.12.1/packages/material-ui-styles/src/makeStyles/makeStyles.d.ts#L12)) makes this unnecessary for consumers upgrading to TS 4.x


ynotdraw (approved):
🙇 thank you!

danperkins (approved):

aecorredor (approved):

PR #170, opened 2021-07-16
fix: Add "align" to Text component

ynotdraw:
### Changes

- Add `align` prop to `<Text`
  - Currently only supporting the three I'm most aware of, `center`, `left`, `right`, and `justify`.  Can add more if folks use other options more frequently. https://developer.mozilla.org/en-US/docs/Web/CSS/text-align

I made this a `fix` conventional commit to do the `PATCH` version.


HaleyWardo (approved):
✨

markdlv (approved):

dexterca (approved):

evanheisler (approved):

smolster (approved):
A perfect PR, docs and all!

PR #173, opened 2021-07-20
fix: checkbox disabled state styles

HaleyWardo:
Here is an [issue](https://github.com/framer/motion/issues/262) that I was able to reference.

**BEFORE**

https://user-images.githubusercontent.com/32574227/126359499-847062ec-ac1a-447c-a77f-76390b179753.mov

**AFTER**

https://user-images.githubusercontent.com/32574227/126359548-19bdbf86-f643-43fa-ae23-74778f0633ee.mov

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -148,7 +148,11 @@ const tickVariants = {
   pressed: (isChecked: boolean) => ({
     pathLength: isChecked ? 0.85 : 0.2,
   }),
-  checked: { pathLength: 1 },
+  checked: {
+    pathLength: 1,
+    stroke: 'var(--tick-primary)',
+    strokeOpacity: 1,
+  },
   unchecked: { pathLength: 0 },
 };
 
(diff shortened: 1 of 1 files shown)


RyanRatcliff (approved):

HaleyWardo:
Right now we are just animating if checked, but if the user disables/enables the checkbox dynamically there is no way to re-render/re-animate. Setting a multi-variant animate will allow us to re-animate if `disabled` or `checked` changes. 

I am setting the variant to `enabled` or `disabled` however we don't have a variant for that. If we truly want to use it as it should be, I think we would strip out the disabled and enabled styles from checked and unchecked and move them into those two properties. But for right now this works... unless you guys think otherwise.

HaleyWardo:
The more I think about it, the more I want to strip those styles out and put them into the enabled and disabled variants. I'm unsure exactly which properties those would be though. I'm guessing it would just be the stroke but I need to look at the styles again.

ynotdraw (approved):

ynotdraw:
> The more I think about it, the more I want to strip those styles out and put them into the enabled and disabled variants. I'm unsure exactly which properties those would be though. I'm guessing it would just be the stroke but I need to look at the styles again.

Yeah, up to you.  I think that may make some sense.  We also may want to think about the `readOnly` state in this case.

dexterca (approved):

HaleyWardo:
> > The more I think about it, the more I want to strip those styles out and put them into the enabled and disabled variants. I'm unsure exactly which properties those would be though. I'm guessing it would just be the stroke but I need to look at the styles again.
> 
> Yeah, up to you. I think that may make some sense. We also may want to think about the `readOnly` state in this case.

Looks like we need to add a readOnly state.

PR #179, opened 2021-08-06
feat: add start and end adornments to TextField

HaleyWardo:
**Changes**
- Added `startAdornment` and `endAdornment` as new props

Screen Shot 2021-08-06 at 7 49 26 PM

Screen Shot 2021-08-08 at 7 42 13 PM

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -140,6 +140,10 @@ export const useStyles = makeStyles(
         userSelect: 'none',
       },
 
+      '& a': {
+        padding: 0,
+      },
+
       '& svg': {
         color: theme.palette.black[500],
         height: theme.pxToRem(20),
@@ -152,15 +156,7 @@ export const useStyles = makeStyles(
       },
     },
     adornmentInverse: {
-      '& p': {
-        color: theme.palette.common.white,
-      },
-
-      '& svg': {
-        color: theme.palette.common.white,
-      },
-
-      '& button': {
+      '& p, & svg, & button': {
         color: theme.palette.common.white,
       },
     },
@@ -230,6 +226,8 @@ export interface TextFieldProps
   label?: BaseFormElement['label'];
   secondaryLabel?: string;
   tooltipMessage?: string;
+  // Adornment Recommended components:
+  // Icon, IconButton, IconButtonLink, or Text
   startAdornment?: React.ReactNode;
   endAdornment?: React.ReactNode;
 }
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -48,6 +48,10 @@ export const useStyles = makeStyles(
         opacity: 0.9,
       },
     },
+    inputContainer: {
+      position: 'relative',
+      width: 'fit-content',
+    },
     input: {
       backgroundColor: 'rgba(132, 137, 166, 0.15)',
       border: `1px solid transparent`,
@@ -77,7 +81,7 @@ export const useStyles = makeStyles(
         '&:focus': {
           backgroundColor: 'rgba(132, 137, 166, 0.15)',
           border: `1px solid transparent`,
-        }
+        },
       },
       '&::-webkit-input-placeholder': {
         color: theme.palette.black[400],
@@ -103,7 +107,7 @@ export const useStyles = makeStyles(
         opacity: 1,
         '&:focus': {
           backgroundColor: 'rgba(230, 231, 237, 0.1)',
-        }
+        },
       },
       '&::-webkit-input-placeholder': {
         color: 'rgba(255, 255, 255, 0.8)',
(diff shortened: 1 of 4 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -174,6 +230,8 @@ export interface TextFieldProps
   label?: BaseFormElement['label'];
   secondaryLabel?: string;
   tooltipMessage?: string;
+  startAdornment?: React.ReactNode;
```
HaleyWardo, line 233: @ynotdraw and I tried typing these to only be specific components (`Icon`, `IconButton` and `Text`) but didn't have any luck.

It looks like[ MUI is typing theirs](https://github.com/mui-org/material-ui/blob/d6ea629e6babc7de5b6a1ccf0effcd360eb6f37f/packages/material-ui/src/InputAdornment/InputAdornment.d.ts#L16) as `React.ReactNode` as well. They just have a comment that states: 
```
    /**
     * The content of the component, normally an `IconButton` or string.
     */
```

I also checked the [Typescript cheatsheet ](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/basic_type_example) and it also says `React.ReactNode`.

In our Storybook docs you can see that I added it is recommended that they use Icon, IconButton, or Text. Does anyone feel like I need to add a comment here as well so if anyone looks at the source code they will see those recommendations? 🤷‍♀️

HaleyWardo:
@dexterca I know we chatted on Friday about you coming in and updating some of the styling. Do you have some time to look at this either today or tomorrow? I can rebase this to a feature branch if so.

ynotdraw (dismissed):

dexterca:
> @dexterca I know we chatted on Friday about you coming in and updating some of the styling. Do you have some time to look at this either today or tomorrow? I can rebase this to a feature branch if so.

Yes, I can take a look today.

xdumaine:
I wonder if we should switch some some of the "special" text input fields like `UnitNumberFormatField` inputs like `PercentFormatField` or `PriceFormatField` to use these instead of putting their prefix/suffix inside the text? Worth experimenting with, at least. cc @dexterca

xdumaine (dismissed):

ynotdraw:
> I wonder if we should switch some some of the "special" text input fields like `UnitNumberFormatField` inputs like `PercentFormatField` or `PriceFormatField` to use these instead of putting their prefix/suffix inside the text? Worth experimenting with, at least. cc @dexterca
> 
> Just experimenting here's what it looks like so you can kind of compare (I'm not sure I like it):
> 
> Today	Adornments
> image	image

Excellent suggestion!  I dig this a lot personally 👍

dexterca:
@xdumaine @ynotdraw After talking thru some things with @HaleyWardo and looking at Xander's examples, I think we're going to limit adornments to `icon`, `iconButton` and `iconButtonLink` on this PR. We'll then rely on `...FormatField`s for text-based "adornments". Thanks Xander for bringing this to our attention!

HaleyWardo:
Converting to draft while @thedoryan works on some styling improvements. 🌟

src/components/TextField/TextField.tsx
```diff
@@ -48,14 +48,25 @@ export const useStyles = makeStyles(
...
+    },
     input: {
       backgroundColor: 'rgba(132, 137, 166, 0.15)',
       border: `1px solid transparent`,
       borderRadius: theme.pxToRem(4),
       color: theme.palette.text.primary,
       fontFamily: theme.typography.fontFamily,
       fontSize: theme.pxToRem(14),
-      minWidth: theme.pxToRem(175),
+      minWidth: theme.pxToRem(180),
```
dexterca, line 69: Not sure if this qualifies as a breaking change? I needed to bump the `minWidth` of `<TextField` to get consistent widths across Chrome, Firefox and Safari. This could impact some layouts?

ynotdraw: Hm I probably wouldn't consider it a breaking change since it's only 5 pixels and there weren't breaking API changes; however, I'm curious what other folks think

src/components/TextField/TextField.tsx
```diff
@@ -92,6 +103,9 @@ export const useStyles = makeStyles(
         color: theme.palette.black[400],
       },
     },
+    inputHasAdornment: {
+      maxWidth: theme.pxToRem(152),
+    },
```
ynotdraw, line 108: I'm curious about the reasoning for this class - below it looks like we are using padding to shift the cursor for the input left / right depending on the adornment positioning (start/end). Are the `padding` attributes not enough in this case where it'll extend the width of the element?

dexterca: Correct, without this `maxWidth`, `<`TextField`s with adornments and without `fullWidth` will have unequal widths as compared to `<TextFields`s without adornments.

Without `maxWidth` widths are uneven
Screen Shot 2021-08-10 at 3 23 44 PM

ynotdraw: Ah nice thanks for the info!

ynotdraw (approved):

xdumaine (approved):

PR #186, opened 2021-08-30
try  a simple change to avoid throwing security error on localstorage

jsking216:


src/utils/canAccessLocalStorage.ts
```diff
@@ -0,0 +1,8 @@
+export const canAccessLocalStorage = () => {
+  try {
+    localStorage.getItem('test');
```
ynotdraw, line 3: Does something like:

```
(typeof window.localStorage !== 'undefined')
```

work as well instead of always `get`ting a fake item from storage?

jsking216: i tried a variation of ^ the undefined check and `'localStorage' in window` as well... it seems that the only option is to catch the access error

src/utils/canAccessLocalStorage.ts
```diff
@@ -0,0 +1,8 @@
...
+    return true;
+  } catch {
+    return false;
+  }
+};
```
ynotdraw, line 8: How would you feel about co-locating this in the `LayoutManager` directory as that's the only place using it at the moment?

src/utils/canAccessLocalStorage.ts
ynotdraw: Ah okay, interesting.  In that case, do you think we should name the key for this something a bit specific to chroma, like `chroma-react-storage` or something instead of `test`?

src/utils/canAccessLocalStorage.ts
jsking216: sure!

src/utils/canAccessLocalStorage.ts
jsking216: sure, the actual string doesn't matter as any kind of usage of localStorage will trigger the security exception

src/components/LayoutManager/LayoutManager.tsx
```diff
@@ -94,7 +95,8 @@ export const LayoutManager: React.FC<LayoutManagerProps> = ({
         }
         setIsSidebarCollapsed((state) => {
           const isOpen = !state;
-          localStorage &&
+          canAccessLocalStorage() &&
+            localStorage &&
```
macnaughton, line 99: on these checks, do we need the `&& localStorage` here if we've already established `canAccessLocalStorage()`?

jsking216: you're right,  `canAccessLocalStorage` will catch that 👍

ynotdraw (approved):
Cool thanks for the PR!

PR #189, opened 2021-09-08
fix: Remove relative position of both radio groups' root (fieldset)

dexterca:
When adding a `RadioGroupMinimal` to my project I noticed the text label (`<legend>`) was missing. Removing `position: relative` from the root (`<fieldset`) made the text label appear. I took a look at both `<RadioGroup` and `<RadioGroupMinimal` and from what I can tell I don't think `position: relative` is needed. Removing relative positioning didn't seem to have any visual changes in stories across Chrome, Firefox and Safari.

@ynotdraw can you confirm if we need relative positioning or not? These components are leveraging `<fieldset>` and `<legend>` so there might be something I'm missing. 

RELATIVE POSITION / LABEL HIDDEN | POSITION UNSET / LABEL VISIBLE
-- | --
Screen Shot 2021-09-08 at 10 18 30 AM | Screen Shot 2021-09-08 at 10 18 37 AM

### DIFF
--- a/src/components/Radio/RadioGroup.tsx
+++ b/src/components/Radio/RadioGroup.tsx
@@ -18,7 +18,6 @@ export const useStyles = makeStyles(
       margin: 0,
       minWidth: 0,
       padding: 0,
-      position: 'relative',
       verticalAlign: 'top',
       width: '100%',
     },
--- a/src/components/Radio/RadioGroupMinimal.tsx
+++ b/src/components/Radio/RadioGroupMinimal.tsx
@@ -18,7 +18,6 @@ export const useStyles = makeStyles(
       margin: 0,
       minWidth: 0,
       padding: 0,
-      position: 'relative',
       verticalAlign: 'top',
       width: '100%',
     },


ynotdraw (approved):
Hey sorry for the delay here.  Yeah, I think this is probably fine.  I don't see any children of the root element relying on `position: 'absolute'`, so unless consumers are doing something strange and absolutely positioning things inside of the radio group, we should be okay 👍 

For a quick sanity check as well, may be able to try yarn linking to PHC and double-check all is good there.  I imagine everything will be just fine.  @HaleyWardo may know where exactly in the app we're using these components.

HaleyWardo:
> Hey sorry for the delay here. Yeah, I think this is probably fine. I don't see any children of the root element relying on `position: 'absolute'`, so unless consumers are doing something strange and absolutely positioning things inside of the radio group, we should be okay 👍
> 
> For a quick sanity check as well, may be able to try yarn linking to PHC and double-check all is good there. I imagine everything will be just fine. @HaleyWardo may know where exactly in the app we're using these components.

I've personally used them in the Insight module settings, Subject Viewer Program Progress module, and Subject Search

dexterca:
Looks like we're good. I found four examples when I yarn linked PHC (2 w/ label and 2 w/o labels). You can see `position: relative` has been removed from `fieldset` in bottom right of every screenshot.

Screen Shot 2021-09-09 at 10 35 46 AM
Screen Shot 2021-09-09 at 10 33 14 AM
Screen Shot 2021-09-09 at 10 16 09 AM
Screen Shot 2021-09-09 at 10 07 15 AM

PR #191, opened 2021-09-28
Upgrade reakit and framer-motion

HaleyWardo:
I noticed several console warnings when running tests in a repo that uses `chroma-react`. The warnings were coming from chroma so I decided to fix them since I am trying to tidy up our console. Seems like the way to fix most of them was to just bump the version for both.

**Error fixed**
- `Uncaught [Error: Non-integer amount of pennies passed to PriceFormatField]`
  - We were throwing an error so it was logging the error to the console 

**Warnings fixed**
-  `It's recommended to have at least one tabbable element inside dialog. The dialog element has been automatically focused. `

- `You are trying to animate stroke from "#FFFFFF" to var(--tick-primary). "#FFFFFF" is not an animatable value - to enable this animation set #FFFFFF to a value animatable to var(--tick-primary) via the style property.`

- `Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

- `Can't focus composite item component because ref wasn't passed to component.`

- `Can't determine if the element is a native tabbable element because ref wasn't passed to the component.`

**UX Change**
- With the latest version of Reakit, the first item in a menu is no longer auto-focused.  This may be considered a breaking change if your tests relied on focus + keyboard usage.


smolster (approved):

src/components/Modal/Modal.tsx
```diff
@@ -234,7 +234,7 @@ const ModalInner = React.forwardRef<HTMLDivElement, ModalProps>(
               }
             })}
             ref={thisRef}
-            positionTransition
+            layout
```
HaleyWardo, line 237: `positionTransition` was replaced with `layout`. https://www.framer.com/docs/guide-upgrade/##layout-animations

src/components/Snackbar/Snackbar.tsx
```diff
@@ -158,7 +158,7 @@ export const Snackbar: React.FC<SnackbarProps> = React.forwardRef<
             role={role}
             onMouseEnter={onMouseEnter}
             onMouseLeave={onMouseLeave}
-            positionTransition
+            layout
```
HaleyWardo, line 161: `positionTransition` was replaced with `layout`. https://www.framer.com/docs/guide-upgrade/##layout-animations

ynotdraw (approved):
Nice, thanks for the improvements 👍

src/components/Tabs/TabPanel.tsx
```diff
@@ -37,6 +37,7 @@ export const TabPanel: React.FC<TabPanelProps> = ({
     <BaseTabPanel
       {...tab}
       className={clsx(classes.root, className)}
+      tabId={stopId}
```
HaleyWardo, line 40: https://github.com/reakit/reakit/blob/master/packages/reakit/CHANGELOG.md#100-2020-05-12

HaleyWardo:
@dexterca if you have some time today, would you mind pulling this down and testing some of the components? I went through and manually tested them but I know you are the framer-motion pro and would like to have a second pair of eyes to make sure nothing is broken. 🙏🏻

dexterca:
I'll pull it down and take a look sometime later today.

dexterca (approved):
Pulled it down, took a look at all the components that import framer-motion and everything LGTM.

HaleyWardo:
I'm tempted to revert these changes as there have been a few issues related and unexpected changes to some components. Has anyone ran into any issues?

PR #193, opened 2021-10-06
feat: Add trailing icon support for button

dexterca:
Adds support for icons to be placed on the right side of buttons

Screen Shot 2021-10-06 at 10 21 15 AM
Screen Shot 2021-10-06 at 10 20 08 AM

### DIFF
--- a/src/components/Button/Button.test.tsx
+++ b/src/components/Button/Button.test.tsx
@@ -1,5 +1,8 @@
 import * as React from 'react';
-import { IconComponent } from '../../testUtils/IconComponent';
+import {
+  IconComponent,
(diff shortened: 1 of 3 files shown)


ynotdraw (approved):
Looks great to me 👍

It'd probably be good to add a deprecation comment above `MenuButton.tsx` saying they can use this instead.  Then eventually we can delete `MenuButton` in favor of this in a breaking change release.

Something like this should be sufficient:

```js
/** @deprecated Please use `trailingIcon` from Button instead. This will be removed in a future release */
export const MenuButton = React.forwardRef<HTMLButtonElement, MenuButtonProps>(
```

PR #211, opened 2021-11-02
fix: Larger radio clickbox

nicolls1:
In reference to thread in slack channel #frontend-dev

### DIFF
--- a/src/components/Radio/Radio.tsx
+++ b/src/components/Radio/Radio.tsx
@@ -11,12 +11,14 @@ import {
 import { generateUniqueId } from '../_private/UniqueId';
 import { Text } from '../Text';
 import { useRadioGroup } from './useRadioGroup';
+import mergeRefs from 'react-merge-refs';
 
 export const RadioStylesKey = 'ChromaRadio';
 
 export const useStyles = makeStyles(
   (theme) => ({
     root: {
+      cursor: 'pointer',
       display: 'flex',
       '& > input[type="radio"]::after': {
         opacity: 0,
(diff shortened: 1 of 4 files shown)


src/components/Radio/Radio.tsx
```diff
@@ -143,7 +146,11 @@ export const Radio = React.forwardRef<HTMLInputElement, RadioProps>(
     };
 
     return (
-      <div className={clsx(classes.root, className)}>
+      <div
+        role="radio"
+        className={clsx(classes.root, className)}
+        onClick={() => inputRef?.current?.click()}
```
rob-snyder, line 152: Does this `onClick` sync to the onChange call in the `<input` tag through the `inputRef`?

rob-snyder (approved):
Looks good to me, though I'd defer to someone else's approval.

src/components/Radio/Radio.tsx
nicolls1: Yup, it clicks the input element which triggers whatever callbacks the input has when clicked. So clicking this parent div is equivalent to clicking the input itself. 

Its common to have a setup like this for file inputs when you don't want to use the native file input display so you hide the file input and click it via ref when some other element is clicked.

src/components/Radio/Radio.tsx
```diff
@@ -143,7 +146,11 @@ export const Radio = React.forwardRef<HTMLInputElement, RadioProps>(
     };
 
     return (
-      <div className={clsx(classes.root, className)}>
+      <div
+        role="radio"
```
ynotdraw, line 150: I'm curious about having this here as well as on the `<input` on how a screen reader will behave? I can dig in a bit here later today

nicolls1: I added it because it was giving an error, but I'm not sure what best practice for this would be.

nicolls1: Digging in would be great!

PR #214, opened 2021-11-03
fix: Add `fullWidth` prop to `<RadioGroupMinimal`

### LINKED ISSUE #213: Add `fullWidth` prop to RadioGroupMinimal
`<RadioGroupMinimal` should not span 100% of parent's width by default. This is more noticeable when there are only two radio options.

Screen Shot 2021-11-03 at 9 55 14 AM

dexterca:
Adding `fullWidth` prop so that by default `<RadioGroupMinimal` doesn't span 100% width of parent. When there was only a couple of radio options, the design looked a little wonky on account of options being spread to extreme left and right when `direction` was set to `row`.

BEFORE | AFTER (default) | AFTER (fullWidth)
-- | -- | --
01-radio-min-before | 02-radio-min-after-default | 03-radio-min-fullWidth

**UPDATED DOCS**

Screen Shot 2021-11-03 at 11 23 42 AM

**UPDATED KNOBS**

`direction="row"` | `direction="column"`
-- | --
knob-row | knob-column

Closes #213

### DIFF
--- a/src/components/Radio/RadioGroupMinimal.tsx
+++ b/src/components/Radio/RadioGroupMinimal.tsx
@@ -12,14 +12,12 @@ export const useStyles = makeStyles(
   (theme) => ({
     root: {
       border: 0,
-      display: 'inline-flex',
       flexDirection: 'column',
       height: '100%',
       margin: 0,
       minWidth: 0,
       padding: 0,
       verticalAlign: 'top',
-      width: '100%',
     },
     legend: {
       color: theme.palette.black[800],
@@ -38,7 +36,7 @@ export const useStyles = makeStyles(
       background: 'rgba(132, 137, 166, 0.15)',
       borderRadius: theme.pxToRem(20),
       border: 'solid 1px transparent',
-      display: 'flex',
+      display: 'inline-flex',
       flexDirection: 'column',
       margin: 0,
       overflow: 'hidden',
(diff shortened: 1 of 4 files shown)


src/components/Radio/RadioGroupMinimal.tsx
```diff
@@ -105,6 +106,11 @@ export const useStyles = makeStyles(
         },
       },
     },
+    radiosFullWidth: {
+      '& > div': {
```
dexterca, line 110: Not sure this is the most elegant solution? 🤷🏻

stories/components/Radio/Radio.stories.tsx
```diff
@@ -57,6 +57,7 @@ const getMinimalPropOptions = (): RadioGroupMinimalProps => {
       },
       'row'
     ),
+    fullWidth: boolean('fullWidth', false),
```
dexterca, line 60: Add `fullWidth` knob

ynotdraw (commented):
Looks great to me! Only suggestion is adding a test to ensure the right classes get applied when adding the `fullWidth` prop.  Let me know if you need a hand!

dexterca:
> Looks great to me! Only suggestion is adding a test to ensure the right classes get applied when adding the `fullWidth` prop. Let me know if you need a hand!

✅

ynotdraw (approved):
🎉

xdumaine (approved):

rob-snyder (approved):

PR #220, opened 2021-11-09
Test coverage cleanups

smolster:
## Motivation
Something weird: the `master` build in GitHub Actions is claiming 85% branch coverage, but my local runs are claiming <80 (causing failure).

Doesn't seem like an urgent issue since the master build is continuously passing, but thought I'd fixup a couple files that were bringing us down.

I fixed two problems:
- No existing test files or storybookfor the `Keymap` component. I added a dead-simple test and a storybook entry.
  - I wonder if this is worth keeping around in Chroma. It adds a dependency, and [there is only one use case in the whole org](https://github.com/search?q=org%3Alifeomic+KeymapHelp&type=code).
- An unused `useWindowSize` hook that I deprecated and `istanbul ignore`d.
  - Why it's dead: there's an identical hook located at a more friendly filepath (`hooks/events/useWindowSize`).

### DIFF
--- a/src/components/KeymapHelp/KeymapHelp.test.tsx
+++ b/src/components/KeymapHelp/KeymapHelp.test.tsx
@@ -1,0 +1,19 @@
+import * as React from 'react';
+
+import { KeymapHelp, KeymapHelpProps } from '.';
+import { renderWithTheme } from '../../testUtils/renderWithTheme';
+
+const LISTENER_TEST_ID = 'key-listener-test-id';
+
+const render = (props?: KeymapHelpProps) =>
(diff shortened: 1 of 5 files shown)


ynotdraw (approved):
Thanks for doing this.

- I definitely agree that `KeyMap` should be moved out of Chroma eventually. It was added as part of a hackathon, but doesn't really meet our standards anymore for when something should go in Chroma versus an app itself where it's being used.
- Excellent find on the window size hook!

xdumaine (approved):

HaleyWardo (approved):

PR #221, opened 2021-11-09
Add an onError handler to avatar img

xdumaine:
### Motivation:

Sometimes, avatar images fail to load, and causes layout issues.

Example: 
image

This `img` src fails to load because it's set to my google profile image, but my cookie for google auth has expired. It causes overflow and horizontal scroll on the whole page. 

### Solution:

Just emit an error to consumer (if they want it) and fallback to initials, hiding the failed `img` tag. 

image

image


smolster (approved):
This change seems okay to me, though it might introduce new behavior to clients (showing initials when an error arises). Probably fine since I'd assume errors rarely happen at this level in most of our apps.

dexterca (approved):
🧹  Thanks for cleaning this up!

PR #222, opened 2021-11-09
fix: allow chromicons v2 as a peerDependency

smolster:
## Motivation
See [this internal Slack thread](https://lifeomic.slack.com/archives/CHMFL4PQT/p1636471415066500).

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,14 +30,14 @@
     }
   },
   "peerDependencies": {
-    "@lifeomic/chromicons": "^1.2.1",
+    "@lifeomic/chromicons": ">=1.2.1",
     "react": "16",
     "react-dom": "16",
     "react-router-dom": ">=5.1.2"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0-0",
-    "@lifeomic/chromicons": "^1.2.1",
+    "@lifeomic/chromicons": "^2.2.0",
     "@material-ui/types": "^5.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -30,14 +30,14 @@
     }
   },
   "peerDependencies": {
-    "@lifeomic/chromicons": "^1.2.1",
+    "@lifeomic/chromicons": ">=1.2.1",
```
sternetj, line 33: Should this one be bumped higher as well?

package.json
```diff
@@ -30,14 +30,14 @@
...
     "react": "16",
     "react-dom": "16",
     "react-router-dom": ">=5.1.2"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0-0",
-    "@lifeomic/chromicons": "^1.2.1",
+    "@lifeomic/chromicons": "^2.2.0",
```
sternetj, line 40: Pr #222 and we're updating it to version 2.2 🤩

xdumaine (approved):

package.json
smolster: If we bump this to e.g >=2.2.0, it will unnecessarily cause warnings for consumers still using v1

sternetj: Ahh okay that makes sense

sternetj (approved):

HaleyWardo (approved):

PR #224, opened 2021-11-10
fix: add anchorPosition prop to DayPicker

smolster:
## Motivation
Adding very basic manual control of anchor positioning in `DayPicker` via an `anchorPosition` prop. Maybe in the future, this component could support elegant auto-positioning based on viewport collisions, etc.

I'm hoping that this will suffice for a while, and that individual consumers who need that kind of elegant behavior can implement it themselves until we have time to add it to Chroma (I know that at least a couple initial consumers can live without fancy automatic adjustments for now).

## Behavior
In order, shows:
- bottom-left (default)
- bottom-right
- top-right (note the poor behavior with viewport collisions -- it's currently on developers to pick a sane direction)
- top-left

### DIFF
--- a/src/components/DayPicker/DayPicker.tsx
+++ b/src/components/DayPicker/DayPicker.tsx
@@ -23,10 +23,12 @@ const setToMidnight = (date: Date) => {
 };
 
 const useStyles = makeStyles((theme) => ({
+  root: {
+    position: 'relative',
+  },
   dayPicker: {
     position: 'absolute',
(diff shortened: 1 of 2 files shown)


ynotdraw (approved):
Nice, I dig your approach here and agree about fancier auto-collision things can come later 🎉

aecorredor (approved):

xdumaine (approved):

PR #226, opened 2021-11-11
fix: add min/max support to DayPicker

smolster:
## Motivation
Two:
1. Adding a mechanism to specify a min/max date, as well as selectively disabling dates arbitrarily.
2. Allow specifying arbitrary props to the internal day picker, to allow more flexible customization.

### DIFF
--- a/src/components/DayPicker/DayPicker.tsx
+++ b/src/components/DayPicker/DayPicker.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 import clsx from 'clsx';
 
 import ReactDayPicker, { LocaleUtils } from 'react-day-picker';
-// Is this safe? consumers might not be using webpack w/ CSS loaders,
 import 'react-day-picker/lib/style.css';
 import { Calendar, ChevronLeft, ChevronRight } from '@lifeomic/chromicons';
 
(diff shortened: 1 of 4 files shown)


src/components/DayPicker/DayPicker.tsx
```diff
@@ -15,11 +14,13 @@ import { composeEventHandlers } from '../../utils';
...
-  date.setMinutes(0);
-  date.setSeconds(0);
-  date.setMilliseconds(0);
+const midnightOf = (date: Date) => {
+  const copy = new Date(date.toISOString());
+  copy.setHours(0);
+  copy.setMinutes(0);
+  copy.setSeconds(0);
+  copy.setMilliseconds(0);
+  return copy;
```
smolster, line 23: immutability is better!

danperkins (approved):
Looks good to me - to prevent thrashing on Chroma versions would it be worth allowing the consumer to specify arbitrary `ReactDayPicker` props and spread those?

smolster:
@danperkins seems like a good idea. I'll target for a subsequent PR.

smolster:
Actually, I'll just add to this PR for ease.

xdumaine (approved):

src/components/DayPicker/DayPicker.tsx
ynotdraw: Great idea 👍

ynotdraw (approved):
Excellent work!

HaleyWardo (approved):

danperkins (approved):

PR #227, opened 2021-11-11
feat: Tab list pill variant + fullWidth

nicolls1:
This PR adds a `pill` variant to tabs. It also adds a `fullWidth` prop to both `default` and `pill` variants. The `pill` variant styling is intended to match the styling of `<RadioGroupMinimal`. Also note, `default` tabs have text that wraps where as `pill` tabs text truncates with overflow ellipsis.

### AFTER

Screen Shot 2021-11-16 at 9 17 17 AM

Screen Shot 2021-11-16 at 9 18 59 AM


src/components/Tabs/TabList.tsx
```diff
@@ -27,18 +34,20 @@ export interface TabListProps {
   className?: string;
 }
 
-export const TabList: React.FC<TabListProps> = ({
+export const TabList: React.FC<TabListProps & Partial<TabListOptions>> = ({
```
nicolls1, line 37: Thought passing the types of the base component for these would be nice. Any concerns with this?

ynotdraw: I _think_ it'd be okay, my only worry may be there could be something we don't want to support from a design perspective (e.g., `orientation`, etc.).

We had initially only exposed the props we approved from a design perspective intentionally ( cc @dexterca ) .  I'm a bit torn personally on exposing this, my gut feeling says leave things as-is for now until we have a case where we specifically want one of these additional props, but I don't have strong opinions either way 😄

src/components/Tabs/TabsContext.tsx
```diff
@@ -1,3 +1,13 @@
...
 
-export const TabsContext = React.createContext<any>({ selectedId: null });
+export interface TabsState {
+  tabState: TabStateReturn;
+  variant: 'default' | 'pill';
+}
+
+export const TabsContext = React.createContext<TabsState>({
+  // only will happen when tabs components are used outside of the context
+  tabState: {} as any,
```
ynotdraw, line 11: Nice, yeah thanks for cleaning this up.  I completely missed this when removing `selectedId` before 🙇

src/components/Tabs/TabList.tsx
nicolls1: Sounds good, lets see if there are any other opinions else we can hold off

nicolls1:
Waiting on actual pill styles to merge

dexterca:
👀 I'm going to take a look at pill styles today. I plan on just adding my changes to this PR unless someone objects to that.

nicolls1:
Checked it out locally and LGTM!

src/components/Tabs/Tab.tsx
```diff
@@ -48,6 +48,61 @@ export const useStyles = makeStyles(
         opacity: 0.4,
       },
     },
+    pill: {
+      borderBottom: 'none',
+      borderRadius: '50px',
```
ynotdraw, line 53: Should this use `theme.pxToRem(50)` instead to be consistent with the `borderRadius` on line 65 and the `borderRadius` set here https://github.com/lifeomic/chroma-react/pull/227/files#diff-0243aba83c9530a47511fd3f1ac4a05bc5dbfafa61918171d15d0f7a6c456994R21?

src/components/Tabs/Tabs.test.tsx
```diff
@@ -21,6 +21,38 @@ test('it renders Tabs', async () => {
...
+        <Tab stopId="tab2">Tab 2</Tab>
+      </TabList>
+    </Tabs>
+  );
+
+  const root = await findByTestId(testId);
+  const options = root.querySelectorAll('[role="tab"]');
+  expect(options).toBeTruthy();
+  expect(options.length).toEqual(2);
+});
```
ynotdraw, line 54: Could this test verify a `className` has `fullWidth` somewhere via `toHaveClass()`?

src/components/Tabs/Tabs.test.tsx
```diff
@@ -21,6 +21,38 @@ test('it renders Tabs', async () => {
...
```
ynotdraw, line 38: Is there a way to check that the `pill` variant is applied somewhere in the DOM?  May be able to leverage `toHaveClass()` and assert on that.

stories/components/Tabs/default.md
```diff
@@ -37,6 +37,15 @@ A class name can be provided.
...
 
+### Full Width
+
+To set the width of tabs to take 100% of the width, use the `fullWidth`
+prop.
+
+```jsx
+<Tabs fullWidth>…</Tabs>
+```
+
```
ynotdraw, line 48: Should the `variant` prop be added to the docs as well for visibility?

src/components/Tabs/Tab.tsx
dexterca: ✅

stories/components/Tabs/default.md
dexterca: ✅

src/components/Tabs/Tabs.test.tsx
dexterca: @nicolls1 can you take a look at updating per @ynotdraw's suggestion?

src/components/Tabs/Tabs.test.tsx
dexterca: @nicolls1 can you take a look at updating per @ynotdraw's suggestion?

nicolls1: Pushed

src/components/Tabs/Tabs.test.tsx
nicolls1: Pushed

ynotdraw (approved):

smolster (approved):
Nice work!

PR #232, opened 2021-12-02
fix: open day picker to currrently selected month by default

nicolls1:
Master build is failing since its a new month and the calendar isn't opening to the selected month by default. I'm guessing this was the expected behavior judging from the tests.


smolster (approved):
Nice catch!

PR #236, opened 2022-01-06
Fix flexWrap on box

xdumaine:
I noticed that the component had `flexWrap: boolean`, which I wanted, but it didn't do anything.

### DIFF
--- a/src/components/Box/Box.tsx
+++ b/src/components/Box/Box.tsx
@@ -83,6 +83,7 @@ export const useStyles = makeStyles<BoxProps>(
       alignBaseline: { alignItems: 'baseline' },
       alignCenter: { alignItems: 'center' },
       alignStart: { alignItems: 'start' },
+      flexWrap: { flexWrap: ({ flexWrap }) => (flexWrap ? 'wrap' : 'nowrap') },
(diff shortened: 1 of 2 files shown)


src/components/Box/Box.tsx
```diff
@@ -83,6 +83,7 @@ export const useStyles = makeStyles<BoxProps>(
       alignBaseline: { alignItems: 'baseline' },
       alignCenter: { alignItems: 'center' },
       alignStart: { alignItems: 'start' },
+      flexWrap: { flexWrap: ({ flexWrap }) => (flexWrap ? 'wrap' : 'nowrap') },
```
ynotdraw, line 86: Should this be:

```
flexWrap: { flexWrap: 'wrap' }
```

When I tried `flexWrap={false}` I was expecting `flex-wrap: nowrap` based on this logic to be applied to the element, but it appears it did not apply.  I could totally be missing something here too

Screen Shot 2022-01-06 at 9 17 09 PM

xdumaine: Yeah, it looks like there's some double logic being applied. I could either switch this to your suggestion, or remove the clsx class to make this always apply. Will update.

ynotdraw (approved):
🎉

PR #243, opened 2022-01-24
feat: Add `gap` prop to `<Box`

### LINKED ISSUE #241: Add `gap` prop to `<Box`
Since [all modern browsers now support flexbox `gap` property](https://caniuse.com/?search=gap), add `gap` support to `<Box`.

dexterca:
Since [all modern browsers now support flexbox `gap` property](https://caniuse.com/?search=gap), add `gap` support in order to more easily control spacing of children within `<Box`. 

### AFTER
https://user-images.githubusercontent.com/485903/150874488-5cdac7de-da1c-49a3-af52-32b60ec4500b.mov

## Why? It’s a nice way to prevent unwanted spacing

Have you ever used margins to create spacing between elements? If we’re not careful, we can end up with extra spacing before and after the group of items.

Solving that usually requires adding negative margins or resorting to pseudo-selectors to remove margin from specific items. But the nice thing about using `gap` in more modern layout methods is that you only have space _between_ items. The extra cruft at the start and end is never an issue!

https://css-tricks.com/almanac/properties/g/gap/

Close #241


src/components/Box/Box.tsx
```diff
@@ -62,6 +63,11 @@ export const useStyles = makeStyles<BoxProps>(
       if (typeof value === 'number') return theme.spacing(value);
     };
 
+    const stringOrThemeSpacingGap = (value: string | number | undefined) => {
+      if (typeof value === 'string') return value;
+      if (typeof value === 'number') return theme.spacing(value) + 'px';
```
HaleyWardo, line 68: I'm not too familiar with this so sorry if I am wrong, but doesn't `theme.spacing` already return the value as pixels?

dexterca: This confused me as well. Using `+ px` was the only way I could get `gap` to work with strings and spacing units. I'm not really sure what the problem is. I can tell you I've had to add `px` when using `gap` with `theme.spacing` in other Lifeology repos as well. 

It looks like MUI had this same problem at one point: https://github.com/mui-org/material-ui/issues/24773

ynotdraw: Ah interesting, yeah, seems like `px` is required here for this: https://github.com/mui-org/material-ui/pull/24794/files#diff-c66b437e5d952a0998e624e32687b4f3114020624f2fee56d4894f27579874f3R281

> Use spacing unit in `gap`, `rowGap`, and `columnGap`. If you were using a number previously, you need to mention the px to bypass the new transformation with `theme.spacing`.

HaleyWardo (approved):

ynotdraw (approved):

src/components/Box/Box.tsx
xdumaine: would `theme.pxToRem` be a better solution than concatenating `px`?

PR #245, opened 2022-01-26
feat: Add flexChildren prop

### LINKED ISSUE #242: Add `flexChildren` prop to `<Box`
Create prop that adds `flex: 1` to direct children of `<Box` so children will flex to full width/height of parent

dexterca:
I find myself declaring `'& > *': { flex: 1 }` a lot with `<Box` to make children fill parent container so I thought of adding this prop as a quick way to accomplish this.

### AFTER

https://user-images.githubusercontent.com/485903/151193473-518ddc0b-46f1-4c5c-94f5-22832ce1350a.mov

flex-children-docs

Closes #242

### DIFF
--- a/src/components/Box/Box.tsx
+++ b/src/components/Box/Box.tsx
@@ -16,6 +16,7 @@ export interface BoxProps
   > {
   ref?: React.Ref<HTMLDivElement>;
   className?: string;
+  flexChildren?: boolean;
   flexWrap?: boolean;
   align?:
     | 'stretch'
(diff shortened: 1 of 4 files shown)


src/components/Box/Box.tsx
```diff
@@ -16,6 +16,7 @@ export interface BoxProps
   > {
   ref?: React.Ref<HTMLDivElement>;
   className?: string;
+  flexChildren?: boolean;
```
xdumaine, line 19: I was going to suggest that this be typed as typeof `flex` but I think we could add that later if there's a need, without it being a breaking change.

xdumaine (approved):

ynotdraw (approved):

PR #247, opened 2022-02-08
docs: Change spacing to correct numbering on doc site

jhbeard:
This PR only add spaces to the beginning of each line. It does not change any text or formatting.

The step numbering currently appears correctly on github, but is reset to all 1s when automatically imported into phc-docs at https://docs.dev.lifeomic.com/development/open-source/chroma-react/. I believe this is due to the python-markdown used by phc-docs interpreting the .md file differently than the GitHub-flavored-Markdown of Github. Specifically, python-markdown thinks the consecutive list is restarting at 1 (despite the 2 or 3 in the text) when there is no spacing at the lines between the numbered steps. Adding two tabs at each line seems to fix this.
<type>: docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,39 +20,39 @@ To get started with Chroma, follow these steps:
 
 1. Install dependencies
 
-```bash
-yarn add @lifeomic/chroma-react @material-ui/styles react-router-dom
-```
+    ```bash
+    yarn add @lifeomic/chroma-react @material-ui/styles react-router-dom
+    ```
 
-Chroma leverages `@material-ui/styles` for CSS-in-JS and `react-router-dom` for link-related components.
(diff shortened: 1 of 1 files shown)


jhbeard:
> Hey @jhbeard this looks great, thanks for doing this! To make our bot happy, you'll need to update your commit message to be something like `docs: Change spacing to correct numbering on doc site`. Probably the easiest way to do this is [to amend your commit](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message). If you have any questions/issues with this, let me know, I'd be happy to help!

Thanks Tony. I will flail around with it a bit, but may reach out if I have problems.

PR #248, opened 2022-02-08
docs: Change spacing to correct numbering on doc site

jhbeard:
This PR only add spaces to the beginning of each line. It does not change any text or formatting. (Repeat of earlier cancelled PR and deleted branch)

The step numbering currently appears correctly on github, but is reset to all 1s when automatically imported into phc-docs at https://docs.dev.lifeomic.com/development/open-source/chroma-react/. I believe this is due to the python-markdown used by phc-docs interpreting the .md file differently than the GitHub-flavored-Markdown of Github. Specifically, python-markdown thinks the consecutive list is restarting at 1 (despite the 2 or 3 in the text) when there is no spacing at the lines between the numbered steps. Adding two tabs at each line seems to fix this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,4 +115,5 @@ Want to override the default theme of Chroma? No problem!
       const classes = useStyles({});
       return <button className={classes.root}>Custom Button</button>;
     };
+    
     ```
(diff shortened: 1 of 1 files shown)


jhbeard:
Hey Tony, I think I fixed the commit message and passed lifeomic-probot. Thanks

Issue #37, opened 2020-11-02
Upgrade to the latest version of Storybook


ynotdraw:
## Summary

A major version of Storybook is out and it has MDX support.  It would be neat to migrate to the latest version and also potentially move to using MDX instead of `storybook-readme`?

nicolls1:
#301

PR #259, opened 2022-04-13
deps: Upgrade jest, typescript, ts-jest, ts-loader

ynotdraw:
### Changes

- Upgrade jest, typescript, ts-jest, ts-loader
  - Pre-work for getting to https://github.com/lifeomic/chroma-react/issues/37
- Make adjustments based on above upgrades

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,5 @@
 module.exports = {
+  testEnvironment: 'jsdom',
   transform: {
     '^.+\\.tsx?$': 'ts-jest',
   },
@@ -14,6 +15,9 @@ module.exports = {
     '<rootDir>/src/hooks/**/*.{ts,tsx,js,jsx}',
     '<rootDir>/src/styles/**/*.{ts,tsx,js,jsx}',
     '!<rootDir>/src/styles/overrides/*.{ts,tsx,js,jsx}',
+    '!<rootDir>/src/components/index.ts',
+    '!<rootDir>/src/styles/index.ts',
+    '!<rootDir>/src/components/**/index.ts',
   ],
   testPathIgnorePatterns: ['/node_modules/', '<rootDir>/dist/'],
   setupFilesAfterEnv: [
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^3.0.0",
     "husky": "^4.2.5",
-    "jest": "^25.4.0",
+    "jest": "^27.5.1",
     "jest-tap-reporter": "^1.9.0",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.1.1",
@@ -83,9 +83,9 @@
     "semantic-release": "^17.4.4",
     "storybook-readme": "^5.0.8",
     "tap-xunit": "^2.4.1",
-    "ts-jest": "^25.4.0",
-    "ts-loader": "^7.0.1",
-    "typescript": "^3.8.3"
+    "ts-jest": "^27.1.4",
+    "ts-loader": "^8.1.0",
+    "typescript": "^4.6.3"
   },
   "dependencies": {
     "@material-ui/core": "^4.9.11",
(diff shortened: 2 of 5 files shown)


jest.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  testEnvironment: 'jsdom',
```
ynotdraw, line 2: Had to add this after upgrading `jest`: https://jestjs.io/docs/configuration#testenvironment-string

> If you are building a web app, you can use a browser-like environment through [jsdom](https://github.com/jsdom/jsdom) instead.

jest.config.js
```diff
@@ -14,6 +15,9 @@ module.exports = {
     '<rootDir>/src/hooks/**/*.{ts,tsx,js,jsx}',
     '<rootDir>/src/styles/**/*.{ts,tsx,js,jsx}',
     '!<rootDir>/src/styles/overrides/*.{ts,tsx,js,jsx}',
+    '!<rootDir>/src/components/index.ts',
+    '!<rootDir>/src/styles/index.ts',
+    '!<rootDir>/src/components/**/index.ts',
```
ynotdraw, line 20: We use the "barrel pattern", where the `index` file is always just a re-export of the components.  So for example:

`src/components/Button`
`src/components/Button/Button.tsx`
`src/components/Button/index.ts`

The `index` file just re-exports functions and components from `Button.tsx`:

```js
export { Button, ButtonProps, ButtonStylesKey, ButtonClasses } from './Button';
```

Due to this pattern, it's safe to remove these from collected coverage imo

package.json
```diff
@@ -83,9 +83,9 @@
     "semantic-release": "^17.4.4",
     "storybook-readme": "^5.0.8",
     "tap-xunit": "^2.4.1",
-    "ts-jest": "^25.4.0",
-    "ts-loader": "^7.0.1",
-    "typescript": "^3.8.3"
+    "ts-jest": "^27.1.4",
+    "ts-loader": "^8.1.0",
```
ynotdraw, line 87: Can't go to `ts-loader` 9 yet as it has a minimum webpack version of `5` (we are on `4`) https://github.com/TypeStrong/ts-loader/releases/tag/v9.0.0.  Moving to the newer version of webpack will be part of the storybook PR.

src/components/Snackbar/Snackbar.test.tsx
```diff
@@ -136,12 +136,17 @@ it('renders a Snackbar with the default isOpen', () => {
   expect(root).toBeNull();
 });
 
-test('it does not schedule the Snacbar to close if it is not already open', () => {
+test('it does not schedule the Snackbar to close if it is not already open', () => {
```
ynotdraw, line 139: Typo with `snackbar` here

src/components/Snackbar/Snackbar.test.tsx
```diff
@@ -136,12 +136,17 @@ it('renders a Snackbar with the default isOpen', () => {
...
   jest.useFakeTimers();
+
   renderWithTheme(<Snackbar duration={1500} isOpen={false} />);
+
+  jest.runAllTimers();
+
   // We can't assert setTimeout wasn't called at all, because the testing
   // library calls it
-  expect(setTimeout).not.toHaveBeenCalledWith(expect.any(Function), 1500);
+  expect(setTimeoutSpy).toBeCalledTimes(0);
```
ynotdraw, line 149: I updated this test to ensure `setTimeout` is not called by using `jest.spyOn` instead.  Not sure why I didn't do that before 🤔

test/mockMuiStyles.js
```diff
@@ -14,9 +14,8 @@ const removeHash = (classes = {}) => {
 };
 
 jest.mock('@material-ui/styles/makeStyles', () => {
-  const makeStylesActual = require.requireActual(
-    '@material-ui/styles/makeStyles'
-  ).default;
+  const makeStylesActual = jest.requireActual('@material-ui/styles/makeStyles')
```
ynotdraw, line 17: `require.requireActual` is no longer a thing as of jestv26 and the recommended path is to use this jest version instead 👍

evanheisler (approved):

smolster (approved):

samirhsnv (approved):

PR #262, opened 2022-05-10
Move slider helper text to always directly follow label

xdumaine:
Before:

Screen Shot 2022-05-10 at 9 19 04 AM

After:

Screen Shot 2022-05-10 at 9 17 34 AM


dexterca (approved):
LGTM ... helper text always displays just below label no matter the label placement.

PR #266, opened 2022-05-19
feat: Add positive and  negative `color` values to `<Button`

dexterca:
Screen Shot 2022-05-19 at 3 01 49 PM
Screen Shot 2022-05-19 at 3 01 10 PM
Screen Shot 2022-05-19 at 3 01 14 PM


src/components/Button/Button.tsx
```diff
@@ -141,6 +141,70 @@ export const useStyles = makeStyles(
...
+      '&:disabled, &[disabled]': {
+        backgroundColor: theme.palette.error.main,
+        color: theme.palette.common.white,
+      },
+    },
+    containedPositive: {
+      backgroundColor: theme.palette.green.main,
+      color: theme.palette.common.white,
+      '&:hover': {
+        backgroundColor: theme.palette.green[900],
```
xdumaine, line 159: Just a thought, but it might be worth introducing a `theme.palette.primary.positive` and `theme.primary.primary.negative` to make it easier to override these colors. Thinking about skillspring, where the primary color is green.

dexterca: ✅ I took a stab at it. I might need some more direction if my latest changes don't provide what you're expecting.

xdumaine (approved):

PR #271, opened 2022-05-20
feat: Add positive and negative `color` values to `<IconButtonLink`

dexterca:
Screen Shot 2022-05-20 at 12 43 46 PM
Screen Shot 2022-05-20 at 12 43 50 PM

No docs update for this PR since we state:
_An element that appears as an Icon Button component, but is a link underneath the covers. It has the same properties as Icon Button exposed._


smolster (approved):

loscm:
Requested review from a random `@lifeomic/security` team member, @bishopb, due to term: **expose**.

PR #273, opened 2022-05-20
feat: add MenuGroupHeadings

smolster:
## Motivation
I've wanted these for a while. Seems others have too.

This PR also introduces minor sizing changes to `Menu`, per feedback from @dexterca.

### DIFF
--- a/src/components/Menu/MenuItem.tsx
+++ b/src/components/Menu/MenuItem.tsx
@@ -14,12 +14,12 @@ export const useStyles = makeStyles(
       cursor: 'pointer',
       display: 'flex',
       fontFamily: theme.typography.fontFamily,
-      fontSize: theme.typography.body1.fontSize,
+      fontSize: theme.pxToRem(14),
       letterSpacing: '0.015625em',
       margin: 0,
       outline: 'none',
       overflow: 'hidden',
-      padding: theme.spacing(1.5, 2.5),
+      padding: theme.spacing(1, 2),
       userSelect: 'none',
       width: '100%',
       '&:hover,&:focus': {
@@ -30,7 +30,7 @@ export const useStyles = makeStyles(
       },
     },
     icon: {
-      marginRight: theme.spacing(2),
+      marginRight: theme.spacing(1),
     },
   }),
   { name: MenuItemStylesKey }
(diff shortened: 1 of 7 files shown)


src/components/Menu/MenuItem.tsx
```diff
@@ -14,12 +14,12 @@ export const useStyles = makeStyles(
       cursor: 'pointer',
       display: 'flex',
       fontFamily: theme.typography.fontFamily,
-      fontSize: theme.typography.body1.fontSize,
+      fontSize: theme.pxToRem(14),
```
mdlavin, line 17: Curious, is there a move away from a fixed list of styles like `body1`?

mdlavin (approved):

src/components/Menu/MenuItem.tsx
dexterca: Tony and I decided awhile back to start using `pxToRem` over the fixed list but I can't remember the exact reason as it's been so long ago. We have a mix of both in Chroma now. I realize this is a terrible answer to your question.

dexterca (approved):

stories/components/Menu/Menu.stories.tsx
```diff
@@ -90,6 +95,76 @@ const MenuStory: React.FC = () => {
...
+          <Menu
+            aria-label="Chroma menu options children"
+            anchorElement={
+              <IconButton
+                aria-label="Click to open children only menu"
+                icon={MoreHorizontal}
+              />
+            }
+            items={[
+              <MenuGroupHeading>First Section</MenuGroupHeading>,
```
aecorredor, line 108: My only thought was this: is there an advantage of not abstracting away `MenuGroupHeading` inside `Menu` via some sort of prop? I thought this because `MenuGroupHeading` is very specific and meant to be used within the `Menu` context. I thought that having the menu take something like the following could work:

```js
sections: [
   { 
       title: 'Header Title', // this would be optional
       items: [],
   }
]
```

smolster: @aecorredor good question. I went this route because there was prior art in the library -- this is very similar to how `Select` works, for example: https://github.com/lifeomic/chroma-react/blob/master/stories/components/Select/Select.stories.tsx#L136-L148

smolster: It was also a non-breaking change, which I felt was important. Your proposed solution could also be done in a non-breaking way -- just mentioning for clarity.

aecorredor: Makes sense!

aecorredor (approved):

PR #275, opened 2022-06-09
feat: Add secondary text to `MenuItem`

dexterca:
Screen Shot 2022-06-09 at 2 16 49 PM
Screen Shot 2022-06-09 at 2 17 09 PM
Screen Shot 2022-06-09 at 2 17 20 PM
Screen Shot 2022-06-09 at 2 17 34 PM
Screen Shot 2022-06-09 at 2 17 52 PM

### DIFF
--- a/src/components/Menu/MenuItem.tsx
+++ b/src/components/Menu/MenuItem.tsx
@@ -2,6 +2,7 @@ import clsx from 'clsx';
 import * as React from 'react';
 import { makeStyles } from '../../styles';
 import { GetClasses } from '../../typeUtils';
+import { Box } from '../Box';
 
 export const MenuItemStylesKey = 'ChromaMenuItem';
 
@@ -20,6 +21,7 @@ export const useStyles = makeStyles(
       outline: 'none',
       overflow: 'hidden',
       padding: theme.spacing(1, 2),
+      textAlign: 'left',
       userSelect: 'none',
       width: '100%',
       '&:hover,&:focus': {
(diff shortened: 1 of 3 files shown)


src/components/Menu/MenuItem.tsx
```diff
@@ -71,7 +99,12 @@ export const MenuItem = React.forwardRef<HTMLButtonElement, MenuItemProps>(
             height={18}
           />
         )}
-        {!!text && text}
+        <Box direction="column" align="flex-start" gap={0.25}>
+          <Box className={classes.text}>{!!text && text}</Box>
+          <Box className={classes.secondaryText}>
+            {!!secondaryText && secondaryText}
+          </Box>
```
HaleyWardo, line 106: Should we render this `div` even if there is no `secondaryText` passed in or should we conditionally render this whole section.

```suggestion
         {!!secondaryText && (
            <Box className={classes.secondaryText}>
               {secondaryText}
            </Box>
          )}
```

HaleyWardo (dismissed):
This looks awesome! Thank you for adding it. I just had one small nit.

src/components/Menu/MenuItem.tsx
MMFane: Agreed on this nit

src/components/Menu/MenuItem.tsx
```diff
@@ -28,9 +30,23 @@ export const useStyles = makeStyles(
...
+        },
+      },
     },
     icon: {
+      alignSelf: 'flex-start',
       marginRight: theme.spacing(1),
+      minWidth: theme.pxToRem(18),
+    },
+    text: {
+      lineHeight: 1.35,
```
MMFane, line 45: Curious - does or could the theme have variables for typical line heights so this doesn't have to be hard-coded here? Guessing it would be ideal for this to change along with any changing font-sizes.

MMFane (approved):
Thanks for giving me a practical first look at Chroma stories! Just a couple nits, otherwise great work. I'm excited to get into Chroma with y'all.

src/components/Menu/MenuItem.tsx
smolster: Adding basic line height standards seems reasonable to me. Also, fine to do in a separate PR.

smolster (approved):
Looks great!

PR #279, opened 2022-06-10
feat: Add negative and positive colors to `<Pill`

dexterca:
Screen Shot 2022-06-10 at 2 48 56 PM
Screen Shot 2022-06-10 at 2 49 02 PM
Screen Shot 2022-06-10 at 2 51 08 PM

### DIFF
--- a/src/components/Pill/Pill.test.tsx
+++ b/src/components/Pill/Pill.test.tsx
@@ -19,12 +19,28 @@ test('it renders a Pill with a label', async () => {
   expect(root).toBeInTheDocument();
 });
 
-test('it renders a Pill with the provided "color"', async () => {
+test('it renders a Pill with primary "color"', async () => {
   const { findByTestId } = renderWithTheme(
     <Pill data-testid={testId} color="primary" />
   );
   const root = await findByTestId(testId);
   expect(root).toHaveClass('ChromaPill-primaryColor');
+});
+
+test('it renders a Pill with negative "color"', async () => {
+  const { findByTestId } = renderWithTheme(
+    <Pill data-testid={testId} color="negative" />
+  );
+  const root = await findByTestId(testId);
+  expect(root).toHaveClass('ChromaPill-negative');
+});
+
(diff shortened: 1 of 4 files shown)


src/components/Pill/Pill.test.tsx
```diff
@@ -19,14 +19,30 @@ test('it renders a Pill with a label', async () => {
   expect(root).toBeInTheDocument();
 });
 
-test('it renders a Pill with the provided "color"', async () => {
```
MMFane, line 22: I'm curious - did you used to be able to define whatever color you wanted and now we're providing a limited set of available colors by concern? I'm new to Chroma - want to make sure I'm on the same page for when I start working in here.

src/components/Pill/Pill.tsx
```diff
@@ -24,6 +24,12 @@ export const useStyles = makeStyles(
     primaryColor: {
       background: theme.palette.primary.main,
     },
+    negative: {
```
MMFane, line 27: Nit. Seeing this for the first time, my thought was "negative" and "positive" would switch the brightness/values of the background and text. Could just be me; I was already in this headspace from another badge PR.

It's supposed to mean "positive" like "pass" or "good" and "negative" as "fail" or "bad", right? I think the names are fine, but am commenting in case enough others feel the same that a rename is warranted.

MMFane (approved):
Looks great! Just a question and a nit.

src/components/Pill/Pill.test.tsx
dexterca: `default` and `primary` were the only color options prior to this. If a custom color is needed, the only way to do that is through a custom class.

src/components/Pill/Pill.tsx
dexterca: You're correct, `positive=good` and `negative=bad`. I preferred this over `success` and `error` because if something is red, it doesn't necessarily mean it's some kind of error.

PR #282, opened 2022-06-21
fix: Use palette for `<DayPicker` background hover color

dexterca:
Use palette for day background hover color vs. default `react-day-picker` color

BEFORE | AFTER
-- | --
Screen Shot 2022-06-21 at 3 26 57 PM | Screen Shot 2022-06-21 at 3 30 01 PM

### Keyboard focus

Screen Shot 2022-06-21 at 3 34 01 PM

### DIFF
--- a/src/components/DayPicker/DayPicker.tsx
+++ b/src/components/DayPicker/DayPicker.tsx
@@ -63,7 +63,7 @@ const useStyles = makeStyles((theme) => ({
       backgroundColor: theme.palette.grey[300],
     },
     '&.DayPicker:not(.DayPicker--interactionDisabled) .DayPicker-Day:not(.DayPicker-Day--disabled):not(.DayPicker-Day--selected):not(.DayPicker-Day--outside):hover': {
-      backgroundColor: fade(theme.palette.primary[50], 0.65),
+      backgroundColor: fade(theme.palette.primary[50], 0.5),
     },
   },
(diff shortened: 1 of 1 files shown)


HaleyWardo:
Honestly, I can hardly see the hover color. At first glance, I thought you removed the color completely. Are we sure this change is accessible?

mstagg (approved):
If accessibility is a factor, just reduce the fade a bit to increase the contrast. Otherwise LGTM.

dexterca:
@HaleyWardo a11y is not really a concern here as it's just a hover state. Keyboard users will get a separate state for keyboard focus (see above screenshot). I'm trying to keep the hover state subtle just like we do elsewhere in the Chrōma (e.g., table row hover).

HaleyWardo:
> @HaleyWardo a11y is not really a concern here as it's just a hover state. Keyboard users will get a separate state for keyboard focus (see above screenshot). I'm trying to keep the hover state subtle just like we do elsewhere in Chrōma (e.g., table row hover). I did go ahead and add a little more contrast but it's more than other hover colors in Chrōma so we've lost some consistency there.

@dexterca I am more worried about the visual impaired (ex. colorblind) users since it was hard for me to even see the blue contrast. As long as the blue contrast meets the WCAG AA guidelines then we should be good!

HaleyWardo (approved):

PR #283, opened 2022-06-22
fix: Add prop to `<DayPicker` for custom selected text color

dexterca:
There's a case in Lifeology where I want to keep the `<DayPicker`'s default background color for selected day because it matches all primary buttons. However, in Lifeology there is not enough contrast between `<DayPicker`'s default background color and the text color (white) of selected day to pass a11y checks. This PR adds a prop `selectedTextColor` to customize  selected day's text color in order pass a11y checks. 

BEFORE | AFTER
-- | --
Screen Shot 2022-06-21 at 8 06 39 PM<br />Fails a11y check | Screen Shot 2022-06-21 at 8 07 19 PM<br />Passes a11y check

Screen Shot 2022-06-21 at 8 25 47 PM

### DIFF
--- a/src/components/DayPicker/DayPicker.tsx
+++ b/src/components/DayPicker/DayPicker.tsx
@@ -122,11 +122,18 @@ export type DayPickerProps = Omit<TextFieldProps, 'value' | 'onChange'> & {
   value?: Date;
 
   /**
-   * A background color to use for the currently selected month.
+   * A background color to use for the currently selected day.
    *
    * Default is the main primary color of the theme.
    */
   selectedBackgroundColor?: string;
+
+  /**
+   * A text color to use for the currently selected day.
+   *
+   * Default is white.
+   */
+  selectedTextColor?: string;
 
   /**
    * Where to anchor the calender pop-up. Default is 'bottom-left'.
@@ -228,6 +235,7 @@ export const DayPicker = React.forwardRef<HTMLInputElement, DayPickerProps>(
     {
       value,
       selectedBackgroundColor,
+      selectedTextColor,
       anchorPosition = 'bottom-left',
       minDate,
       maxDate,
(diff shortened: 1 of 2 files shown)


src/components/DayPicker/DayPicker.tsx
```diff
@@ -122,12 +122,19 @@ export type DayPickerProps = Omit<TextFieldProps, 'value' | 'onChange'> & {
...
    * Default is the main primary color of the theme.
    */
   selectedBackgroundColor?: string;
 
+  /**
+   * A text color to use for the currently selected day.
+   *
+   * Default is white.
+   */
+  selectedTextColor?: string;
```
HaleyWardo, line 136: I think we should control this text color so we alway ensure it is accessible. On `palette` there is a helper function `getContrastText` which returns the accessible text color based on the passed in color (in this case the `selectedBackgroundColor`). If you want an example, I used this in `ColorPicker`.

HaleyWardo (changes requested):

loscm:
Requested review from a random `@lifeomic/security` team member, @PlayDohDoh, due to term: **access**.

src/components/DayPicker/DayPicker.tsx
dexterca: Not sure how this would work because `selectedBackgroundColor` can be any custom value and I believe `getContrastText` only works with `palette` colors? Or are you suggesting we do a breaking change and lock down `selectedBackgroundColor` to a `palette` color where we no longer allow custom values?

HaleyWardo: Yeah the more I think about this, the more I think this goes against our normal patterns. In all of our components we limit the colors to primary or secondary palette colors. I would recommend we introduce a breaking change and replace `selectedBackgroundColor` with `color` ("primary" | "secondary" | "default"). 

I checked all of our implementations of DayPicker and none of them are passing in the `selectedBackgroundColor`. So this change should not cause any breaking changes with our current implementations. 

What are your thoughts?

PR #284, opened 2022-06-22
feat: Add `DayPicker` selected day background color from palette options

dexterca:
This PR locks down `<DayPicker`'s `pickerColor` to two palette options: `primary` (default) and `secondary`. Text color of `pickerColor` for each option will now pull from `getContrastText`.

`pickerColor` is replacing `selectedBackgroundColor`.

AFTER (PRIMARY) | AFTER (SECONDARY)
-- | --
Screen Shot 2022-06-22 at 11 54 42 AM | Screen Shot 2022-06-22 at 11 55 07 AM

Screen Shot 2022-06-22 at 11 55 28 AM

### DIFF
--- a/src/components/DayPicker/DayPicker.tsx
+++ b/src/components/DayPicker/DayPicker.tsx
@@ -126,7 +126,7 @@ export type DayPickerProps = Omit<TextFieldProps, 'value' | 'onChange'> & {
    *
    * Default is the main primary color of the theme.
    */
-  selectedBackgroundColor?: 'primary' | 'secondary';
+  color?: 'primary' | 'secondary';
 
   /**
    * Where to anchor the calender pop-up. Default is 'bottom-left'.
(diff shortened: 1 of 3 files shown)


HaleyWardo:
Thank you for making these changes to follow more of our patterns! 🙌 🙇‍♀️

HaleyWardo (dismissed):

src/components/DayPicker/DayPicker.tsx
```diff
@@ -126,7 +126,7 @@ export type DayPickerProps = Omit<TextFieldProps, 'value' | 'onChange'> & {
    *
    * Default is the main primary color of the theme.
    */
-  selectedBackgroundColor?: string;
+  pickerColor?: 'primary' | 'secondary';
```
smolster, line 129: Can we update the JSDoc on this prop to be more explicit?

Something like `Default is 'primary'`

smolster (changes requested):
This is a breaking change, since we're renaming a prop. We need to add a new commit (or modify an existing commit) that uses a breaking change syntax:

```
feat!: some change
    ^ this makes it a breaking change
```

smolster (dismissed):
Awesome!

smolster (approved):

src/components/DayPicker/DayPicker.tsx
dexterca: ✅

PR #286, opened 2022-06-30
fix: radio group minimal accessibility

HaleyWardo:
**Issue**
- User is not able to tab to focus checked radio 
- User is not able to use arrows to toggle between options

**Changes**
- Updated radio input to include screen reader styles instead of not being displayed at all

**BEFORE**

https://user-images.githubusercontent.com/32574227/176694680-67d34846-0bb6-4589-9141-302a2383b86e.mov

**AFTER**

https://user-images.githubusercontent.com/32574227/176694636-45e97059-49ae-4c4d-997b-a7cd91e1493d.mov


xdumaine (approved):

smolster (approved):
Nice!

PR #287, opened 2022-07-01
Fix stepper accessibility 

### LINKED ISSUE #44: Review Stepper component accessibility
Some research is needed with the Stepper component to see what is expected of these types of components and what gaps need filled when it comes to accessibility.

https://github.com/lifeomic/chroma-react/pull/43#pullrequestreview-543119239

HaleyWardo:
Closes #44 

**Issue**
- Visual impaired users have no way of know which step is the active step
- There are no focus styles when tabbing (we only have hover styles)
- Steps as `div` have no tabIndex so they are missed in the tab order

**Changes**
- Fixed syntax to be more semantic (`ol` & `li`) 
- Added focus styles to steps
- Added `tabIndex` to active `div` steps
  - Button steps already have a `tabIndex` with the button element
- Added `aria-label` to read active step index of out total number of steps
- Added `aria-describedBy` so screen readers can get title, subTitle, and subTitlePill values

**BEFORE**

https://user-images.githubusercontent.com/32574227/176944194-9135f19c-ca4b-464c-a36a-0dc948ea35bc.mov

**AFTER**

https://user-images.githubusercontent.com/32574227/176944173-af03c65f-dd6e-43a0-b25d-eccdfa76dd2c.mov

https://user-images.githubusercontent.com/32574227/176944179-5624733b-3203-4e88-a0e6-2213c972c54f.mov

### DIFF
--- a/src/components/Stepper/Step.tsx
+++ b/src/components/Stepper/Step.tsx
@@ -25,7 +25,7 @@ export const useStyles = makeStyles(
         outline: 'none',
       },
 
-      '&:hover': {
+      '&:hover, &:focus': {
         '&>div': {
           transform: 'scale(1.1)',
           transition: '0.25s cubic-bezier(0.2, 0.8, 0.2, 1)',
@@ -35,6 +35,15 @@ export const useStyles = makeStyles(
     divRoot: {
       minWidth: theme.pxToRem(110),
       padding: theme.spacing(0, 0.875, 1.5, 0.875),
+
+      '&:focus': {
+        outline: 'none',
+
+        '&>div': {
+          transform: 'scale(1.1)',
+          transition: '0.25s cubic-bezier(0.2, 0.8, 0.2, 1)',
+        },
+      },
     },
     activeRoot: {
       borderBottom: `2px solid ${theme.palette.primary.main}`,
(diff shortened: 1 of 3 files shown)


src/components/Stepper/Step.tsx
```diff
@@ -139,19 +150,37 @@ export const Step: React.FC<StepProps> = ({
...
+  index = 0,
+  numberOfSteps,
   onClick,
   subTitle,
   subTitlePillLabel,
   title,
   ...rootProps
 }) => {
   const classes = useStyles({});
+  const ariaLabel = `Step ${index + 1} of ${numberOfSteps}`;
```
smolster, line 162: Should numberOfSteps be required? It seems like otherwise, this will read "Step X of undefined"

HaleyWardo: The `numberOfSteps` prop is being passed in via the `Stepper` component when cloning the provided `Step`. We don't want to require this prop because then it would require all current implementations to pass it in. This is a prop that Chroma is controlling. Hopefully that makes sense.

HaleyWardo: We do the same with the `index` prop.

smolster: Does make sense -- thanks for the explanation.

smolster (approved):

xdumaine (approved):

loscm:
Requested review from a random `@lifeomic/security` team member, @cfbevan, due to term: **access**.

PR #288, opened 2022-07-01
Chrom 120 Add Hex to rgba util function

MMFane:
# What Was Changed
- Added hex to rgba util function and helpers
  - it handles 3 and 6-length hex codes
  - it handles lowercase and capital hex codes
  - it handles opacity in decimal and % format
  - it assumes opacity 1 if opacity arg is left out

## Sanity Checks
- Added tests and saw them fail then pass

## Questions for the Team
- Handling both decimal and % opacity the way I did doesn't allow for 1% opacity in the % format. Would it be better just to pick one format and make people use it?
- Haley sent me a [StackOverflow example](https://stackoverflow.com/questions/21646738/convert-hex-to-rgba), but I chose to go more verbose for readability. Do you think I should simplify anything for performance reasons?


src/styles/utils/colorManipulator.tsx
```diff
@@ -0,0 +1,17 @@
+import { hexToRgb } from '@material-ui/core';
```
xdumaine, line 1: Should this file still be included? Looks like you changed it `.ts` and made changes above

src/styles/createTheme.ts
```diff
@@ -46,6 +47,7 @@ export const createTheme = ({
     typography: createTypography(typography),
     boxShadows: createBoxShadows(boxShadows),
     pxToRem: (size: number) => `${size / 16}rem`,
+    hexToRgba: hexToRgba,
```
HaleyWardo, line 50: Just an FYI, you can also use the shorthand since the key and function are the same name. 

```suggestion
    hexToRgba,
```

SchaeStewart (approved):

HaleyWardo (approved):
This looks awesome! 🥳

HaleyWardo:
>Haley sent me a [StackOverflow example](https://stackoverflow.com/questions/21646738/convert-hex-to-rgba), but I chose to go more verbose for readability. Do you think I should simplify anything for performance reasons?

The stackOverflow was more for guidance and a good place to start. I think what you wrote is perfect! It is easy to read and follow along with. 👏 I don't see any performance issue since it is such a small utility function.  Now if we were looping over several items and doing some crazy manipulation, I would recommend we do something different - but what you have is good!

src/styles/createTheme.ts
MMFane: Ah, thank you! I'll update that if I get back in here; I do usually prefer the shorthand

PR #292, opened 2022-07-23
CHROM 199 Update rgba to hexToRgba()

MMFane:
# What Was Changed

- Replaced all usage of rgba(x, y, z, a) with hexToRgba('some color', a)
  - This allows us to use theme colors from various palettes to keep things consistent and easier to change
  - I went ahead and changed some additional hex colors to use palettes
  - This PR can be an example to anyone who wants to start using the hexToRgba util in other repos

## Sanity Checks

- I searched for rgba() in VSCode
  - I left test files alone, as well as boxshadows since they were declared outside of a component. If you'd like me to address these as well, I can.
  - let me know if I missed any refs!

### DIFF
--- a/src/components/Skeleton/Skeleton.tsx
+++ b/src/components/Skeleton/Skeleton.tsx
@@ -41,8 +41,7 @@ export const useStyles = makeStyles(
 
       '&::after': {
         animation: '$shine 1.6s linear 0.5s infinite',
-        background:
-          'linear-gradient(90deg, transparent, rgba(230, 231, 237, 1), transparent)', // graphite[200] with opacity
+        background: `linear-gradient(90deg, transparent, ${theme.palette.graphite[100]}, transparent)`,
         bottom: 0,
         content: `''`,
         left: 0,
--- a/stories/storyComponents/LabeledInput.tsx
+++ b/stories/storyComponents/LabeledInput.tsx
@@ -1,6 +1,30 @@
 import * as React from 'react';
-import black from '../../src/colors/black';
-import { spacing } from '../storyStyles';
+import { makeStyles } from '../../src/styles';
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    marginBottom: theme.spacing(4),
+  },
+  input: {
+    appearance: 'none',
+    borderRadius: '0.25rem',
+    border: `1px solid ${theme.palette.black[100]}`,
+    paddingTop: '0.5rem',
+    paddingBottom: '0.5rem',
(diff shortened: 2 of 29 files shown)


src/components/Skeleton/Skeleton.tsx
```diff
@@ -41,8 +41,7 @@ export const useStyles = makeStyles(
 
       '&::after': {
         animation: '$shine 1.6s linear 0.5s infinite',
-        background:
-          'linear-gradient(90deg, transparent, rgba(230, 231, 237, 1), transparent)', // graphite[200] with opacity
+        background: `linear-gradient(90deg, transparent, ${theme.palette.graphite[100]}, transparent)`,
```
HaleyWardo, line 44: From the comment, it looks like it should be `theme.hexToRgba(theme.palette.graphite[200], 1)`, is that no longer the case?

stories/storyComponents/LabeledInput.tsx
```diff
@@ -1,6 +1,30 @@
...
+    paddingRight: '0.75rem',
+    boxShadow: `
+      0 1px 3px 0 ${theme.hexToRgba(theme.palette.common.black, 0.1)}, 
+      0 1px 2px 0 ${theme.hexToRgba(theme.palette.common.black, 0.06)}
+    `,
+  },
+  label: {
+    display: 'block',
+    marginBottom: theme.spacing(2),
+    fontWeight: 'bold',
```
HaleyWardo, line 24: Should we use the theme font weight?
```suggestion
     fontWeight: theme.typography.fontWeightBold,
```

HaleyWardo (changes requested):
Nice! Thank you for updating all of these! 

Just had a few comments.

src/components/Skeleton/Skeleton.tsx
MMFane: I can change it to that if you prefer! Here's why it is the way it is: 

- (I just doublechecked) Those values are #e6e7ed, or graphite[100], so I'm guessing it was changed or mislabelled before
- With an alpha of 1, I decided to skip the theme.hexToRgba() part. But if we think this may need alpha in the future and want to decrease cognitive load on that future developer, I'd be happy to change it back.

stories/storyComponents/LabeledInput.tsx
MMFane: Sounds good to me!

swain (approved):

HaleyWardo (approved):

PR #293, opened 2022-07-29
PW-96 Add Required Styling to TextField

MMFane:
# What Was Changed

- Added a left-aligned red asterisk to TextField to show that a field is required

## Questions for the Team

- @HaleyWardo pointed out that the 'required' prop may be confused for validation. Can I get thoughts on a solution?
  - Idea: Change to "styleRequired" or similar
  - Idea: Add docs note that this is just visual and validation still needs to be done

# Screenshots

## Before
Screen Shot 2022-07-29 at 12 41 25 PM

## After
Screen Shot 2022-07-29 at 12 41 09 PM

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -288,7 +288,7 @@ export const TextField = React.forwardRef<HTMLInputElement, TextFieldProps>(
                   color === 'inverse' && classes.requiredInverse
                 )}
               >
-                *
+                &#42;
               </span>
             )}
             {label}
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -234,6 +234,8 @@ export interface TextFieldProps
   // Icon, IconButton, or IconButtonLink
   startAdornment?: React.ReactNode;
   endAdornment?: React.ReactNode;
+  /** This property shows the required asterisk (*). Required validation needs to be implemented separately. */
+  showRequiredLabel?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


loscm:
Requested review from a random `@lifeomic/security` team member, @bishopb, due to term: **risk**.

HaleyWardo:
My suggestion is that the prop be called `showRequiredLabel` or something along that - but I am curious as to what others think. As you mentioned in the description, `required` makes me think that this is a controlled input and validation would be handled. I think we need the prop name to be more specific.

xdumaine:
I've definitely thought that adding `required` would do validation in the past, so good calling it out.

> Idea: Add docs note that this is just visual and validation still needs to be done

If you do this, just do it as a TSDoc on the property so it shows when doing autocomplete. TSDoc supports markdown in most editors so you can make it a bold node.

MMFane:
> I've definitely thought that adding `required` would do validation in the past, so good calling it out.
> 
> > Idea: Add docs note that this is just visual and validation still needs to be done
> 
> If you do this, just do it as a TSDoc on the property so it shows when doing autocomplete. TSDoc supports markdown in most editors so you can make it a bold node.
> 
> Maybe _both_ a different prop name, and the TSDoc description.

Thanks Xander and Haley - I'll do both. I have a suspicion only documenting it will lead to some folks in a hurry missing this detail and might cause unexpected issues for them.

swain (approved):
Code LGTM. I'll leave design feedback to others :)

HaleyWardo (approved):

xdumaine (approved):

PR #295, opened 2022-08-03
CHROM-194 Add All Number Format Subcomponents to Storybook

MMFane:
# What Was Changed

- Phone number was already in there - added Unit, Percent, and Price
- Grouped them all under "Number Format Field"
- Added all unique props to each story
- Referenced and linked to the components each extend or implement
- Had to do some rebasing (_one day I will remember to prefix Chroma commits with \<tag\>:_ ). Let me know if anything looks funky.

# Screenshots

## Before
Screen Shot 2022-08-03 at 11 44 59 AM

## After
Screen Shot 2022-08-03 at 11 45 12 AM
Screen Shot 2022-08-03 at 11 45 26 AM


src/components/Button/Button.tsx
```diff
@@ -71,8 +71,7 @@ export const useStyles = makeStyles(
         boxShadow: theme.boxShadows.focusVisibleInverse,
       },
       '&:disabled, &[disabled]': {
-        backgroundColor: 'rgba(255, 255, 255, 0.45)',
-        opacity: 1,
+        backgroundColor: theme.hexToRgba(theme.palette.common.white, 0.45),
```
xdumaine, line 74: Were these intentional in this PR?

HaleyWardo:
Yeah it looks like your rebase didn't work as expected. I am seeing a bunch of `rgba` changes.

src/components/Button/Button.tsx
MMFane: Hmm... no, I don't think so. That may be an artifact of my rebase. Looking into it

MMFane:
> Yeah it looks like your rebase didn't work as expected. I am seeing a bunch of `rgba` changes.

Damn, no kidding! Let's try that again

src/components/Button/Button.tsx
MMFane: @xdumaine @HaleyWardo OK, fixed it!

xdumaine (approved):

PR #300, opened 2022-08-15
fix: MUI component overrides and styling injection

HaleyWardo:
**Changes**
- Updated component props and styling to previous values so that no breaking changes are introduced
- Added `StyledEngineProvider` export and updated documentation
  - This provider is required in order for our styles to override the MUI emotion component styles

### DIFF
--- a/stories/styles/color/color.stories.tsx
+++ b/stories/styles/color/color.stories.tsx
@@ -15,7 +15,6 @@ const {
   background,
   divider,
   text,
-  type,
   ...additionalColors
 } = palette;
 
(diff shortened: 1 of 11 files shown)


namick (approved):

stories/styles/color/color.stories.tsx
```diff
@@ -15,7 +15,6 @@ const {
   background,
   divider,
   text,
-  type,
```
MMFane, line 18: Is this `type` that's been replaced with `mode`? If so, just curious why you removed it rather than replacing it - are we not currently using it in any stories, perhaps?

MMFane (approved):
👏

stories/styles/color/color.stories.tsx
HaleyWardo: Yeah! I don't know if you remember in our pairing, but this was never being used.

PR #301, opened 2022-08-18
Storybook v6

nicolls1:


src/components/Checkbox/Checkbox.tsx
```diff
@@ -314,6 +314,9 @@ export type CheckboxProps = BaseFormElementWithNodeLabel &
     indeterminate?: boolean;
   };
 
+/**
+ * A checkbox component for form usage. Under the covers, this is an input element with `type="checkbox"`.
```
nicolls1, line 318: Instead of using mdx we can just add comments in the code and it will pick them out and add it to the storybook. In general, I think is self explanatory and we don't need as much text.

package.json
```diff
@@ -36,14 +36,19 @@
...
-    "@babel/core": "^7.0.0-0",
+    "@babel/core": "^7.18.10",
     "@lifeomic/chromicons": "^2.2.0",
-    "@storybook/addon-a11y": "^5.3.18",
-    "@storybook/addon-actions": "^5.3.18",
+    "@storybook/addon-a11y": "^6.5.10",
+    "@storybook/addon-actions": "^6.5.10",
+    "@storybook/addon-essentials": "^6.5.10",
+    "@storybook/addon-interactions": "^6.5.10",
     "@storybook/addon-knobs": "^5.3.18",
```
MMFane, line 45: If we're not using knobs anymore, should this import be removed?

MMFane (approved):
Looks good to me! I'm excited to have to write less for each story

package.json
nicolls1: Good call, I'm still running the storybook with the old components included, will remove at the end

nicolls1:
The correct semantic commit for this would be `docs` correct?

xdumaine (approved):
This will need another approver after me since I merged a small fix for the build.

HaleyWardo (approved):
Wow! The docs look way better 🤩 ! Thank you for making this huge updat!

PR #312, opened 2022-11-10
[CHROM-200] Add Sticky Columns to TableModule

MMFane:
# What Was Changed

- Completed Haley's sticky column start 
  - Sticky column styling
  - Set sticky columns' left values so they don't overlap
  - Made a Sticky story in storybook

## Questions for the Team

- @HaleyWardo @dexterca, I think it's unlikely that real use cases will make much use of having sticky columns not be consecutive, but the solution does offer that ability. I've styled the columns in a way I think makes sense for both consecutive and non-consecutive columns, but I'm curious to hear your thoughts on it as well. **- Approved by @dexterca** 

# Screen Recordings

## Before

https://user-images.githubusercontent.com/5824697/201166263-6edbb46f-c9fe-49aa-8e0b-adbf3223704b.mov

## After

https://user-images.githubusercontent.com/5824697/201166311-1de9f6bc-c1fa-4123-8080-57dd3afe2d22.mov


src/components/TableModule/TableHeaderCell.tsx
```diff
@@ -80,6 +80,13 @@ export const useStyles = makeStyles(
     rotatedIcon: {
       transform: 'rotate(180deg)',
     },
+    isSticky: {
+      background: theme.palette.graphite[50],
+      position: 'sticky',
+      left: 0,
+      zIndex: 4,
+      willChange: 'transform',
```
HaleyWardo, line 88: TIL 🎉

HaleyWardo (commented):
Nice job! I will let others review since I was a contributor on this.

HaleyWardo (commented):
Should we add more tests? Like checking that the last sticky column gets the correct className?

dexterca (dismissed):
Visuals look great! Nice work @MMFane!

MMFane:
> Should we add more tests? Like checking that the last sticky column gets the correct className?

Totally agree. I have checks testing for the sticky class and the last sticky class now. Do you think we need any more?

xdumaine:
jw - have you smoke tested this in firefox/safari? it seems like something that could behave oddly in different browsers?

src/components/TableModule/TableHeaderCell.tsx
```diff
@@ -80,6 +80,13 @@ export const useStyles = makeStyles(
     rotatedIcon: {
       transform: 'rotate(180deg)',
     },
+    isSticky: {
+      background: theme.palette.graphite[50],
+      position: 'sticky',
+      left: 0,
+      zIndex: 4,
```
xdumaine, line 87: how'd you arrive at this? Magic z-index numbers can be problematic. One way to at least help with this is to create a constants file with them, so they're easy to reference around a project, and exported for consumers.

src/components/TableModule/TableHeaderCell.tsx
```diff
@@ -130,7 +141,9 @@ export const TableHeaderCell: React.FC<TableHeaderCellProps> = ({
         ((!header.align && headingsCount > 1 && index === headingsCount - 1) ||
           header?.align === 'right') &&
           classes.rootAlignRight,
-        header.className
+        header.className,
+        isSticky && classes.isSticky,
+        isSticky && 'sticky-cell-hook'
```
xdumaine, line 146: what is `sticky-cell-hook` used for? Just tests? It might be better to use a `data-*` attribute like `data-sticky` for example.

MMFane: It's also used to apply the isStickyLast class to the last sticky cell in each row to form the last column styling with the border (see the useEffect on line 323 of TableModule.tsx). I did this since we can't directly select the MUI isSticky class since its ID changes between builds.

Now I'm realizing I got around this with regex in the tests - which method do you think would be better?

MMFane:
> jw - have you smoke tested this in firefox/safari? it seems like something that could behave oddly in different browsers.

@xdumaine Ooh, I have not. That is a great point. On it!

MMFane:
> jw - have you smoke tested this in firefox/safari? it seems like something that could behave oddly in different browsers.

@xdumaine OK, I did some manual sanity checks and there seem to be no browser-related issues, but this did happen to uncover an issue with the left values staying the same while resizing the window, so I fixed that!

src/components/TableModule/TableHeaderCell.tsx
MMFane: I'm going to fix this in another PR, so I can also move the Modal and Select Z-indexes

xdumaine (dismissed):

src/components/TableModule/TableHeaderCell.tsx
xdumaine: personally I'd go with what you did for isStickyLast, just for consistency, or else I'd use a testid/data attribute, but honestly it's not a major deal either way.

xdumaine (approved):

PR #313, opened 2022-11-15
[NO-JIRA] Add Z-Indicies to Theme

MMFane:
# What Was Changed

- Rather than continuing to hard-code z-indicies in Chroma, let's put them all in one util object
- Added existing z-index values to named properties
- Added a general 0 - 20 object for more one-off use
  - Set this up to be expanded or for additional general objects to be created with a helper function

## Sanity Checks
- Manually spot-checked to make sure values remained the same and were coming through from the util


jkdowdle (approved):

src/styles/utils/zIndex.ts
```diff
@@ -0,0 +1,23 @@
+const zIndex = {
```
HaleyWardo, line 1: Should we update the `zIndex` values in the theme and use those instead of private utils?

HaleyWardo (changes requested):
Quick question

namick (approved):

src/styles/utils/zIndex.ts
HaleyWardo: Only concern is that apps are already using the default MUI `zIndex` and we may cause some breaking changes. We'd just need to check.

MMFane: Good catch! I missed the preexisting theme zIndex, so let me see if I can find it now...

MMFane: OK, I've moved zIndex to theme so we're using MUI's zIndicies as well as our new ones. Let me know if I need to improve any of those changes!

jkdowdle (approved):

src/styles/createZIndex.ts
```diff
@@ -0,0 +1,39 @@
+import muiZIndex from '@mui/material/styles/zIndex';
```
HaleyWardo, line 1: I think we can get rid of this import since we have the same import below. 

```suggestion
```

HaleyWardo (changes requested):

MMFane:
Welp, I broke testing. Looking into it

src/styles/createZIndex.ts
MMFane: For the rest of the team's understanding - Haley and I met and talked about this import (it was also causing issues in the tests). For now, we've decided to copy the zIndex values from MUI because 

1. We timeboxed trying to get the import to work with Jest
2. This will allow us to keep MUI's values at this point in time, even if they update in the future, and avoid breaking changes in zIndex

If it ever makes sense to, we can revisit this issue in the future

HaleyWardo (approved):

PR #315, opened 2022-11-28
[NO-JIRA] Improve TableModule sticky useEffects and ref warning

MMFane:
# What Was Changed
PR in `phc-ui` is failing tests in GitHub, but not in my local. There are a few issues, but one is I was throwing an error for all tables without a ref, when only the tables with sticky columns need the ref.
- Changed the error to a warning
- Added checks so all sticky `useEffects` only run for tables with sticky columns

### DIFF
--- a/src/components/TableModule/TableModule.tsx
+++ b/src/components/TableModule/TableModule.tsx
@@ -316,14 +316,16 @@ export const TableModule = React.memo(
       }, [headings, sortState.sortDirection, sortState.sortKey]);
 
       React.useEffect(() => {
-        const allStickyCells = Array.from(
-          document.querySelectorAll('.sticky-cell-hook')
-        );
-        allStickyCells.forEach((cell, index) => {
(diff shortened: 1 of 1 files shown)


src/components/TableModule/TableModule.tsx
```diff
@@ -316,14 +316,16 @@ export const TableModule = React.memo(
...
       React.useEffect(() => {
-        const allStickyCells = Array.from(
-          document.querySelectorAll('.sticky-cell-hook')
-        );
-        allStickyCells.forEach((cell, index) => {
-          if ((index + 1) % stickyCols.length === 0) {
-            cell?.classList.add(classes.isStickyLast);
-          }
-        });
+        if (stickyCols.length > 0) {
```
xdumaine, line 319: nit: reduce nesting by bailing early in the else case (`if (stickyCols.length === 0) { return; }`)

xdumaine (approved):

PR #316, opened 2022-11-29
chore: move to node 16 and upgrade semantic-release

swain:
## Motivation
Upgrading `semantic-release` to address this dependabot PR: #276.

In order to upgrade, we need to upgrade Node. Went to 16.


MMFane (approved):
🔥

PR #319, opened 2022-11-30
[CHROM-208] Add support for three font families

dexterca:
**Changes**

three-fonts

It's common for designers to use a primary font and then a secondary accent font (e.g., a different font for headings). This PR adds support for a secondary font family. I also wanted to leave the door open for a third font family in case a monospace font is needed in addition to primary and secondary fonts.

It's also possible that two sans serif fonts could be used as primary and secondary typefaces so I opted to move away from naming the first two values by font type (i.e., `sansSerif` or `serif`) in favor of `primary`, `secondary` and `monospace`

I linked this to PHC and breaking changes were easy to resolve. [Here is a good first pass at those changes](https://github.com/lifeomic/phc-ui/compare/TESTING-three-fonts?expand=1).

**Breaking changes**

1. `sansSerif` is now `primary`
```diff
-     sansSerif: CSSProperties['fontFamily'];
+     primary: CSSProperties['fontFamily'];
```
```diff
-     fontFamily: fontFamilies.sansSerif,
+     fontFamily: fontFamilies.primary,
```


HaleyWardo:
Also, piggybacking off of @xdumaine. I do also like `theme.typography.fontFamily.primary` rather than `theme.typography.fontFamilyPrimary`. It follows our other patterns in typography.

dexterca:
After further review and offline discussion with @HaleyWardo and @MMFane, we're going to stick with `theme.typography.fontFamilyPrimary` convention. The gist of it is that font weights follow this same syntax. Instead of making font family and font weight match the color convention, we're opting to keep the existing naming convention for font weights and families.

src/styles/createTypography.ts
```diff
@@ -76,7 +81,8 @@ export interface TypographyOptions
     MUITypographyOptions {}
 
 const baseTypography: TypographyOptions = {
-  fontFamily: fontFamilies.sansSerif,
```
HaleyWardo, line 79: Wait, don't we still need to override MUI's `fontFamily`. I believe removing this will include MUIs font family. I am doing this AFK so may be wrong.

HaleyWardo (changes requested):

src/styles/createTypography.ts
xdumaine: This might be a good reason to avoid the breaking change and keep primary as just `fontFamily` without the `primary` suffix.

dexterca: Ok, I think we're in a good spot now. We need to override MUI `fontFamily` so `fontFamily` will serve as Chroma's primary font. Thanks @HaleyWardo for all your help! 🙇🏻

xdumaine (approved):

HaleyWardo (approved):
Thanks Casey!

PR #320, opened 2022-12-01
[NO-JIRA] Make TableModule Row Menu Sticky Again

MMFane:
# Problem
- Previous changes made it so you had to scroll to the end of a table's columns to access the hoverable context menu on a row
- The desired effect is to see this context menu at all times, even on tables with tons of columns

BEFORE | AFTER
-- | --
table-actions-before | table-action-after

# What Caused the Problem
- Removal of a class, `sticky`, that made the menu stick to the right of the table
- Since there were both `sticky` and `isSticky` classes, I took `sticky` for being redundant and removed it

# Solution
- Added class back to TableModule and TableModuleRow
- Renamed it `stickyMenuButton` to eliminate future confusion

### DIFF
--- a/src/components/TableModule/TableModule.tsx
+++ b/src/components/TableModule/TableModule.tsx
@@ -188,6 +188,11 @@ export const useStyles = makeStyles(
         cursor: 'pointer',
       },
     },
+    stickyMenuButton: {
+      position: 'sticky',
+      willChange: 'transform',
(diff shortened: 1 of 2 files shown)


dexterca (approved):

dexterca:
Thank you!

xdumaine (approved):

PR #326, opened 2023-01-29
[NO-JIRA] Add sticky column hover styling

MMFane:
# What Was Changed

- Noticed sticky table cells lacked hover styling, so added it

# Screenshots

## Before

https://user-images.githubusercontent.com/5824697/215345764-2e2a0f12-00fb-4975-bcac-15db80a778c9.mov

## After

https://user-images.githubusercontent.com/5824697/215345767-4a91beb8-a511-4eb7-a2c5-4d604ce8f4fe.mov


dexterca (approved):
Nice, thanks for picking this up!

HaleyWardo:
Should this be considered a `feat` or a `fix`? I would prefer fix but what do you think?

MMFane:
> Should this be considered a `feat` or a `fix`? I would prefer fix but what do you think?

Good call - I've gotten a little `feat` happy after having that last release I was involved in fail because of semantic commit name issues. Will `fix` do it for a patch release?

HaleyWardo:
> > Should this be considered a `feat` or a `fix`? I would prefer fix but what do you think?
> 
> Good call - I've gotten a little `feat` happy after having that last release I was involved in fail because of semantic commit name issues. Will `fix` do it for a patch release?

Yep a `fix` would be a patch release.

dexterca (approved):

PR #332, opened 2023-03-22
ReactTableModule

shawnzhu:
#331

PR #337, opened 2023-03-24
fix(PHC-4380): avoid break change in type

shawnzhu:
## Before

After merging of #333, type `TableConfiguration<Item>` will require one type argument, which will break all applications that uses `TableModule`.

## After

type `TableConfiguration<Item>` will not require any type argument, it could use `TableConfiguration<RowSelectionRow>` when using the row selection API.

### DIFF
--- a/src/components/TableModule/types.ts
+++ b/src/components/TableModule/types.ts
@@ -39,7 +39,7 @@ export interface TableCell<Item = any> extends TableAlignOptions {
   className?: string;
 }
 
-export interface TableConfiguration<Item extends RowSelectionRow> {
+export interface TableConfiguration<Item = any> {
   header: TableHeader;
   cell: TableCell<Item>;
   isSticky?: boolean;
@@ -53,8 +53,8 @@ export interface TableState {
 }
 
 export interface RowSelectionRow {
-  getIsSelected?: () => boolean;
-  getCanSelect?: () => boolean;
-  toggleSelected?: (value?: boolean) => void;
-  getToggleSelectedHandler?: () => (event: unknown) => void;
+  getIsSelected: () => boolean;
+  getCanSelect: () => boolean;
+  toggleSelected: (value?: boolean) => void;
+  getToggleSelectedHandler: () => (event: unknown) => void;
 }
--- a/src/components/TableModule/TableModule.stories.tsx
+++ b/src/components/TableModule/TableModule.stories.tsx
@@ -3,6 +3,7 @@ import { ComponentStory, ComponentMeta } from '@storybook/react';
 
 import { TableModule } from './TableModule';
 import {
+  RowSelectionRow,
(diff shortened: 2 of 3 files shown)


src/components/TableModule/types.ts
```diff
@@ -39,7 +39,7 @@ export interface TableCell<Item = any> extends TableAlignOptions {
   className?: string;
 }
 
-export interface TableConfiguration<Item extends RowSelectionRow> {
+export interface TableConfiguration<Item = any> {
```
simons5593, line 42: Making the default to be `any` seems too accommodating IMHO. I feel like if you explicitly set `any` yourself, then that's on you, but if the _default_ is `any` then it's on us that's promoting less than ideal typing. What do you think?

src/components/TableModule/types.ts
```diff
@@ -53,8 +53,8 @@ export interface TableState {
 }
 
 export interface RowSelectionRow {
-  getIsSelected?: () => boolean;
-  getCanSelect?: () => boolean;
-  toggleSelected?: (value?: boolean) => void;
-  getToggleSelectedHandler?: () => (event: unknown) => void;
+  getIsSelected: () => boolean;
```
simons5593, line 56: If the prior allows conditional, does this mean that they are all now required?

simons5593 (commented):
My 2c.

src/components/TableModule/types.ts
shawnzhu: I do want to make the type required. however, this PR is about backward compatibility to keep the type requirement as it was in v4.0.2:

https://github.com/lifeomic/chroma-react/blob/2c5973db869a67f1cede15ea2663f79032314dfb/src/components/TableModule/types.ts#L42-L46

I tried to fix the types in #333 but realized that it costs more to fix types than adopting a well-documented one via #332

src/components/TableModule/types.ts
shawnzhu: Yes. When using row selection feature, the table configuration MUST be in type `TableConfiguration< RowSelectionRow>`. E.g., https://github.com/lifeomic/chroma-react/blob/84d1c5fe42995c7c24b53c99e98d4674ed6d7043/src/components/TableModule/TableModule.stories.tsx#L457-L471

When not using row selection feature, the table configuration could be in type `TableConfiguration<any>` just what it is today.

src/components/TableModule/types.ts
simons5593: Gotcha. Makes sense.

src/components/TableModule/types.ts
simons5593: Love it! 👍🏽

simons5593 (approved):
Thanks for the info on my questions! LGTM!

PR #338, opened 2023-03-28
fix: Disable row handler when clicking on a checkbox

shawnzhu:
This seems a known problem of `TableModule` where click event on checkbox in a cell will also trigger row click.

## Before

row selection event will trigger row click

## After

row selection via checkbox will not trigger row click


nicolls1 (approved):
Nice!

Issue #331, opened 2023-03-20
Refactor TableModule with TanStack Table


shawnzhu:
Checkout [TanStack Table](https://tanstack.com/table/v8). 

This is a follow up of #330 to cover:

* adding react-table as implementation and make it backward compatible.
* when using `config` of TableModule, no changes to existing apps that use it.
* when using `columns: ColumnRef[]` to configure TableModule , it will use `react-table` 's implementation.

shawnzhu:
I spent hours trying to conquer the sorting feature then realized that the sorting in `TableModule` is a very special API coupling with custom types and a single hard coded hook (sort). I tried to adapt it with the TanStack table API but it will be LOTS of work. 

Action: I will use the feature toggle route to add TanStack API as optional path of `TableModule` the deprecate the existing implementation over time.

shawnzhu:
#333 is a temporary solution before fully adopt TanStack table API

PR #339, opened 2023-03-28
fix(PHC-4380): achieves backward compatibility #331

shawnzhu:
This will support backward compatibility where table row will be kept as they were when row selection is disabled (default option).


MMFane (approved):
🚀

PR #340, opened 2023-04-22
[PHC-4098] Add secondaryAction to SelectOption

MMFane:
# What Was Changed

- Added new properties to `<SelectOption />` to allow for a 'secondaryAction' so that you can both select an option's value in the parent `<Select />` `onChange` and do something else with the individual `<SelectOption />` (e.g. navigate somewhere with more info about the option)
- Update Select story so it actually selects the value when you click an option

## Questions for the Team

- @dexterca, when you get back into the office, can I get your opinion on the design of this? We can add a link as `children` to a <SelectOption /> as it is, but this takes a lot more vertical space and imo is a little harder to tell which option each link goes with at first glance. Thoughts?

| *As-is option* | *Example new option from this PR* |
|---|---|
| Screenshot 2023-04-20 at 5 05 43 PM | Screenshot 2023-04-21 at 2 16 43 PM |

# Screenshots

## Before

https://user-images.githubusercontent.com/5824697/233761493-b535bc4e-0078-4cd2-b688-7db3824f1a59.mov

## After First Pass

https://user-images.githubusercontent.com/5824697/233761497-8c09bbde-b886-4082-9058-7d88e1d62921.mov

## After Updates

https://user-images.githubusercontent.com/5824697/234093129-e7529ef1-5028-4e5a-9b17-4da494163e69.mov

### DIFF
--- a/src/components/Select/Select.stories.tsx
+++ b/src/components/Select/Select.stories.tsx
@@ -3,7 +3,7 @@ import { ComponentStory, ComponentMeta } from '@storybook/react';
 
 import { Select } from './Select';
 import { SelectOption } from './SelectOption';
-import { Edit } from '@lifeomic/chromicons';
+import { ChevronRight, Edit } from '@lifeomic/chromicons';
 import { GroupHeading } from './GroupHeading';
 
 export default {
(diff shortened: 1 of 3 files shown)


dexterca:
I prefer the new option. 

- The `ChevronRight` icon is a little confusing but it looks like you've built it to where any icon can be used for `secondaryAction`.
- Can we add a tooltip to the icon button?
- The scale of the icon button is too large. This is most noticeable when `secondaryText` is not present and `secondaryAction` is present.

MMFane:
> * The `ChevronRight` icon is a little confusing but it looks like you've built it to where any icon can be used for `secondaryAction`.

Totally fair; the chevron doesn't carry any implied meaning. Do you have a preferred default icon?

> * Can we add a tooltip to the icon button?

Yep, we can! There is one issue - when the `<Select />` opens, the first item is automatically keyboard selected for a11y. This makes the first tooltip pop up in a weird place, I'm guessing because it renders before the first item is fully placed in the document flow. I could attempt to hack a solution with a delay - does that sound good?

> * The scale of the icon button is too large. This is most noticeable when `secondaryText` is not present and `secondaryAction` is present.

Thanks for pointing this out. I shrank the icon and centered it with the text content. How's it look now?

SchaeStewart (approved):
Definitely not the expert here, but it looks good to me

nicolls1 (approved):

PR #341, opened 2023-04-24
PHC-4300 Fixes TypeError when rowData has read-only property

shawnzhu:
## Problem

Since `row` is in `any` type, when it has a read-only property like `__typename` when the row data comes from a GQL response object, the `Object.assign()` will fail. E.g.,

```
TypeError: Cannot assign to read only property '__typename' of object '#<Object>'
        at Function.assign (<anonymous>)
```


MMFane (approved):
💥

simons5593 (approved):
💯 LGTM!

PR #343, opened 2023-04-27
Adjust sticky column border-color, -width and background colors

dexterca:
I wanted to lighten sticky cell background in order to make disabled checkboxes more visible. This also enabled me to use `palette.divider` color for sticky right border color. I also doubled the width of the right-border from `2px` to `4px` to help with delineation between sticky and static.

Another win for using `palette.divider` is we now have a continuous vertical line. Whereas with `primary.main` we had a disconnected blue divider due to the use of two different colors.

Screenshot 2023-04-27 at 2 09 51 PM

BEFORE | AFTER
-- | --
Screenshot 2023-04-27 at 2 02 16 PM | Screenshot 2023-04-27 at 1 57 29 PM


MMFane (approved):
Looks great!

PR #344, opened 2023-05-02
fix(TableModule): provide rowIndex to rowActions

shawnzhu:
Add an optional prop `rowIndex` to `rowActions` to be able to delete current row from table


MMFane (approved):
This is great!

PR #355, opened 2023-08-04
Fix table ref warn

MMFane:
# What Was Changed

- Finally fixing some console pollution I introduced back when we added sticky columns to `<TableModule />`
- We now only show the 'Table doesn't have a ref...' warning if some sticky columns are set
    - This will also prevent unnecessary spacing calculations being run when there are no sticky cols
- We also now only show this warning in Dev

# Screenshots

## Before

This warning appeared for every table on every page that did not have a ref, regardless of whether it needed sticky cols

Screenshot 2023-08-04 at 4 52 22 PM

## After

Now nothing should appear for non-sticky tables, or in Prod

### DIFF
--- a/src/components/TableModule/TableModule.tsx
+++ b/src/components/TableModule/TableModule.tsx
@@ -378,8 +378,9 @@ export const TableModule = React.memo(
           });
           setStickyCellsLeft(stickyCellsLeft);
         } else {
-          console.warn(
-            "Table's forwardRef is null, please set it if you want the sticky cell's left value to be set correctly"
+          warning(
+            process.env.NODE_ENV === 'development',
+            "Chroma Warning: Table's forwardRef is null, please set it in order to set the sticky cell's left value correctly"
           );
         }
       }, [forwardedRef, stickyCols]);
--- a/src/components/TableModule/TableModule.tsx
+++ b/src/components/TableModule/TableModule.tsx
@@ -289,9 +289,7 @@ export const TableModule = React.memo(
       const classes = useStyles({});
 
       warning(
-        Boolean(onRowClick) &&
-          !rowClickLabel &&
-          process.env.NODE_ENV === 'development',
+        Boolean(onRowClick) && !rowClickLabel,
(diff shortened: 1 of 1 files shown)


src/components/TableModule/TableModule.tsx
```diff
@@ -374,8 +378,9 @@ export const TableModule = React.memo(
           });
           setStickyCellsLeft(stickyCellsLeft);
         } else {
-          console.warn(
-            "Table's forwardRef is null, please set it if you want the sticky cell's left value to be set correctly"
+          warning(
+            process.env.NODE_ENV === 'development',
+            "Chroma Warning: Table's forwardRef is null, please set it in order to set the sticky cell's left value correctly"
```
shawnzhu, line 383: `process` object is only available in node.js runtime. it will report error when interpreting it in a browser. So it could just keep these lines as they were

epeters3:
> Finally fixing some console pollution I introduced back when we added sticky columns to <TableModule />

So you're the celebrity that introduced sticky columns! I've been enjoying that feature in our Patient ML Runs table.

src/components/TableModule/TableModule.tsx
MMFane: OK - I was modeling this warning off the one on line 291 - should I adjust that to a plain console.warn as well? I think this util is common around Chrōma, so I'm wondering if we have a larger issue to address.

MMFane:
> > Finally fixing some console pollution I introduced back when we added sticky columns to
> 
> So you're the celebrity that introduced sticky columns! I've been enjoying that feature in our Patient ML Runs table.

Yay! Glad to hear it's been helpful! ✨✨✨

src/components/TableModule/TableModule.tsx
shawnzhu: just use `console.warn()` like what it was. I don't think you need to change these lines except your fix

swain: (I also agree with Shawn's immediate feedback: I'd suggest skipping the `process.env` check and just doing `console.warn`)

MMFane: Cool, thanks for the feedback, guys!

swain (approved):

PR #359, opened 2023-08-10
Add `columnsWidth` prop

dexterca:
### AFTER

Screenshot 2023-08-10 at 2 11 34 PM

---

**NOTE**: This PR is merging into `feat-description-list` branch

### DIFF
--- a/src/components/DescriptionList/DescriptionList.tsx
+++ b/src/components/DescriptionList/DescriptionList.tsx
@@ -17,6 +17,10 @@ export const useStyles = makeStyles<DescriptionListProps>(
       if (typeof value === 'number') return theme.spacing(value);
     };
 
+    const string = (value: string | undefined) => {
+      if (typeof value === 'string') return value;
+    };
+
     return {
       title: {
         color: theme.palette.text.primary,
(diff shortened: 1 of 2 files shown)


src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -17,6 +17,10 @@ export const useStyles = makeStyles<DescriptionListProps>(
       if (typeof value === 'number') return theme.spacing(value);
     };
 
+    const string = (value: string | undefined) => {
+      if (typeof value === 'string') return value;
+    };
+
```
dexterca, line 23: I wasn't sure if this was the proper way to do this, so please let me know if there is a more elegant way.

namick: Depends what you are trying to do.  As written, this function takes a `string` or `undefined` so you can pretty much count on that (otherwise a TypeScript error will surface it).

However, the output of this function checks to make sure it is a `string` and returns it if so.  Otherwise it will be `undefined`, which is what a function returns when it ends without encountering a return statement.  So this function actually doesn't do anything.

Since line 110 defines `columsWidth` as a `string` or `undefined`, I think you don't even need this function and can just trust TypeScript.

So maybe you want to just delete this function and change line 97 to:
```
gridTemplateColumns: ({ columnsWidth }) => columnsWidth,
```

dexterca: ✅ 🙇🏻

shawnzhu (approved):

PR #369, opened 2023-08-17
Add Node 18.x support

MMFane:
Looking forward to other repos moving to Node `18.x`, we're making sure Chrōma can support it as well.

## Notes to the Team
- I wasn't 100% certain what to do with the semantic commit for this one. It's not a breaking change and doesn't change anything for our public API, so I'm thinking it should be a patch. Is there a better tag than `fix:` for patches in this case?

### DIFF
--- a/.github/workflows/pr-branch-build.yml
+++ b/.github/workflows/pr-branch-build.yml
@@ -8,11 +8,14 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16.x, 18.x]
     steps:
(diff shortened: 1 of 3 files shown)


shawnzhu (approved):
Thanks!

Just realized that `docs:build` task is not included during PR build. but it works for me when running `yarn run docs:build` with node 18.17.1

MMFane:
> Thanks!
> 
> Just realized that `docs:build` task is not included during PR build. but it works for me when running `yarn run docs:build` with node 18.17.1

Should I create a new PR to add that in? Is that a necessary step for our new docs work on Chrōma?

PR #371, opened 2023-08-22
build: includes docs:build target into branch build workflow

shawnzhu:
## Before

it doesn't build storybook during PR and branch build workflow

## After

it will exercise the target `docs:build` during PR and branch build workflow.

Although it adds 30s more to the build time, it will make sure the storybook setup works all the time

### DIFF
--- a/.github/workflows/pr-branch-build.yml
+++ b/.github/workflows/pr-branch-build.yml
@@ -10,16 +10,31 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16.x, 18.x]
+        include:
+          - group: 1
+            node-version: 16.x
+            run: |
+              yarn test --ci --reporters="github-actions"
+              yarn build
+          - group: 2
+            node-version: 18.x
+            run: |
+              yarn test --ci --reporters="github-actions"
+              yarn build
+          - group: 3
+            node-version: 18.x
+            run: |
+              yarn run docs:build
     steps:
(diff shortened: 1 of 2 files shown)


shawnzhu:
WoW, it takes 122s to run `yarn run docs:build` in github actions. could be optimized as a separate workflow

swain (approved):
Makes sense to me. 30s seems worth it, for sure.

shawnzhu:
> 30s seems worth it, for sure.

https://github.com/lifeomic/chroma-react/pull/371#issuecomment-1688170250 showed that it is `122s` more in github actions 😓

MMFane:
> 

Yeah, adding another 2 minutes feels borderline to me. With an overall build less than 10 min that's not _terrible_ but maybe that's just because I'm coming from the perspective of working with `phc-ui`. 

@shawnzhu if you were to add this to a separate action, how would that work? Like, would it run concurrently with the other build actions?

shawnzhu:
> would it run concurrently with the other build actions?

Yes. I plan to make it in a separate build job via matrix build so that it still manages the PR build in a single workflow

shawnzhu:
After https://github.com/lifeomic/chroma-react/pull/371/commits/c0f10a486aa28d7edb3b8fbd82fbdb920b6c0a59 it will only increase billable time while the duration of PR check was not changed at all since doc build is faster than that of unit test.

swain (approved):
Fancy!

PR #373, opened 2023-08-22
Update B- Components to Storybook CSF3

MMFane:
# What Was Changed

- Replaced now-defunct CSF 1 & 2 types `ComponentStory` and `ComponentMeta` with `StoryObj` and `Meta` respectively
- Updated all stories to the new single-const format 
- Removed titles, since those are now created from the component name unless you need to override them
    - except for `Box`, which is organized into the Layout group
- Created Story type and added it to all stories for increased type safety
-  Added Stories for Outlined and Text variants on `Button` so we have examples of those properties on at least one button component
- Increased height on `ButtonFloat` stories wrapper so it doesn't get cut off by outer window

# Screenshots

## New Button Stories

Screenshot 2023-08-22 at 4 21 18 PM

## ButtonFloat Height Fix

| Before | After |
| --- | --- |
| Screenshot 2023-08-22 at 4 22 03 PM | Screenshot 2023-08-22 at 4 21 48 PM |

### DIFF
--- a/src/components/ButtonFloat/ButtonFloat.stories.tsx
+++ b/src/components/ButtonFloat/ButtonFloat.stories.tsx
@@ -1,11 +1,10 @@
 import React from 'react';
-import { ComponentStory, ComponentMeta } from '@storybook/react';
+import { StoryObj, StoryFn, Meta } from '@storybook/react';
 
 import { ButtonFloat } from './ButtonFloat';
 import { Edit } from '@lifeomic/chromicons';
 
-export default {
-  title: 'Components/ButtonFloat',
+const meta: Meta<typeof ButtonFloat> = {
   component: ButtonFloat,
   argTypes: {
     disabled: {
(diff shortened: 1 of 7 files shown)


src/components/ButtonFloat/ButtonFloat.stories.tsx
```diff
@@ -18,66 +17,77 @@ export default {
...
-          justifyContent: 'center',
-          alignItems: 'center',
-          top: 0,
-          right: 0,
-          bottom: 0,
-          left: 0,
-        }}
-      >
-        {story()}
+      <div style={{ width: '100%', height: '50px' }}>
```
shawnzhu, line 20: is this the fix to avoid the button float to be cut by other box?

MMFane: Yep!

shawnzhu (approved):

PR #377, opened 2023-08-23
Update DescriptionList Components to Storybook CSF3

MMFane:
# What Was Changed

- Replaced now-defunct CSF 1 & 2 types `ComponentStory` and `ComponentMeta` with `StoryObj` and `Meta` respectively
- Updated all stories to the new single-const format 
- Removed titles from ungrouped components whose name is the same as their filename
- Created Story type and added it to all stories for increased type safety

- @dexterca, this has caused the stories for `DescriptionList` as the parent to ungroup. I think this makes more semantic sense, but what's your opinion? I could see it burying the other story folders for our users as well, which would be a bummer. (See screenshot below)
- @dexterca, Second question: do you think the stories for the sub-components are helpful or could we do without them? I noticed the `-Divider`, `-Details`, and `-Terms` stories look very similar, and I'm not sure if the `-Divider` one is illustrating the divider at all. Since you just made this component, I'm curious to learn your thinking behind making these stories.

Screenshot 2023-08-23 at 2 12 04 PM

# Screenshots

_There should be no visual changes to these components_


dexterca:
> @dexterca, this has caused the stories for DescriptionList as the parent to ungroup. I think this makes more semantic sense, but what's your opinion? I could see it burying the other story folders for our users as well, which would be a bummer. 

@MMFane I favor the way you have things setup here. True, some items might get buried but there are accordions to help with that. Another added benefit is that users can more quickly jump to see specific examples in the side nav because you've removed one level of nesting in the side nav.

dexterca:
> @dexterca, Second question: do you think the stories for the sub-components are helpful or could we do without them? I noticed the -Divider, -Details, and -Terms stories look very similar, and I'm not sure if the -Divider one is illustrating the divider at all. Since you just made this component, I'm curious to learn your thinking behind making these stories.

@MMFane If I'm being honest, I didn't put much thought into this. I was just following the convention we had for `<List`. Stripping out the stories works for me because I'm not sure how much value they actually provide.

dexterca (approved):

MMFane:
> > @dexterca, Second question: do you think the stories for the sub-components are helpful or could we do without them? I noticed the -Divider, -Details, and -Terms stories look very similar, and I'm not sure if the -Divider one is illustrating the divider at all. Since you just made this component, I'm curious to learn your thinking behind making these stories.
> 
> @MMFane If I'm being honest, I didn't put much thought into this. I was just following the convention we had for `<List`. Stripping out the stories works for me because I'm not sure how much value they actually provide.

@dexterca I'm going to strip them out then - my thinking is we put any subcomponent stories we need under the parent and skipping boilerplate subcomponent ones will save us story maintenance time and effort

dexterca (approved):

PR #384, opened 2023-08-28
fix: Adjust layout of `<SlideOver` children and add text overflow for panel title

dexterca:
This PR adjusts the layout for children of `<Header` in `<SlideOver` and also adds text overflow ellipsis for the panel title.

### Changes

- Adjust layout for `<Header` children
- Add min-height and -width so `titleIcon` does not shrink

BEFORE | AFTER
-- | --
Screenshot 2023-08-28 at 10 45 18 AM | Screenshot 2023-08-28 at 10 44 20 AM


MMFane (approved):
✨

PR #389, opened 2023-08-30
feat: Add background-color, padding, margin, height and width props to `<Stepper`

dexterca:
BEFORE | AFTER
-- | --
screencapture-lifeomic-github-io-chroma-react-2023-08-30-13_37_06 | screencapture-localhost-9001-2023-08-30-13_37_16


src/components/Box/Box.tsx
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { makeStyles } from '../../styles';
 import { GetClasses } from '../../typeUtils';
 import { warning } from '../../utils';
-import colorOptions from './colorOptions';
+import colorOptions from '../../colors/colorOptions';
```
dexterca, line 6: I wasn't sure if this is the right location to store `colorOptions` in, but I wanted to keep it DRY and share between `<Box` and `<Stepper` 🤷🏻‍♂️

epeters3 (approved):
LGTM

src/components/Box/Box.tsx
MMFane: Works for me! We can always move it when content or context changes in the future.

MMFane (approved):

PR #399, opened 2023-09-15
Add support for tooltip message in `<ComboBox`

dexterca:
Screenshot 2023-09-15 at 12 04 03 PM

### DIFF
--- a/src/components/Select/ComboBox.test.tsx
+++ b/src/components/Select/ComboBox.test.tsx
@@ -4,6 +4,7 @@ import { fireEvent, waitFor, within } from '@testing-library/dom';
 import { renderWithTheme } from '../../testUtils/renderWithTheme';
 import * as React from 'react';
 import { press } from 'reakit-test-utils';
+import { IconComponent } from '../../testUtils/IconComponent';
 
 const testId = 'ComboBox';
 const optionId = 'SelectOption';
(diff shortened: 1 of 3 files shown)


src/components/Select/ComboBox.test.tsx
```diff
@@ -413,3 +414,18 @@ test('it renders an inverse color * when the field is required', async () => {
...
+  const { findByRole } = renderWithTheme(
+    <ComboBox
+      {...props}
+      icon={IconComponent}
+      tooltipMessage="tooltipMessage"
+      data-testid={testId}
+    />
+  );
+
+  const icon = await findByRole('img', { hidden: true });
```
dexterca, line 429: `findByRole('img'` is failing here because there are two SVGs in ComboBox. 

@sternetj @shawnzhu @sam Any ideas on how to be more specific when finding the tooltip icon?

sternetj: I haven't tried it, but based on the [docs](https://testing-library.com/docs/queries/byrole), you might be able to filter by the container. Something like:

```typescript
await findByRole(await findByTestId('ComboBox'), 'img', { hidden: true })
```

dexterca: @sternetj I couldn't get that to work 🤷‍♂️

shawnzhu: the `IconComponent` has a testId `iconComponent`, so you can do:

```TSX
import {
  IconComponent,
  testId as iconTestId,
} from '../../testUtils/IconComponent';
...
  const { findByTestId } = renderWithTheme(
    <ComboBox
      {...props}
      icon={IconComponent}
      tooltipMessage="tooltipMessage"
      data-testid={testId}
    />
  );

  const icon = await findByTestId(iconTestId);
```

shawnzhu: I like this conversation started by the new jest github actions reporter!

dexterca: ✅ 🙏🏻

shawnzhu (approved):

sternetj (approved):

PR #409, opened 2023-10-18
feat: Add custom classname for icon and trailing icon in `<ListItem`

dexterca:
Adds support for adding custom classnames to icon and trailing icon in  `<ListItem`


MMFane (approved):
✨

PR #410, opened 2023-10-19
feat: Add secondary header to `<ExpansionPanel`

dexterca:
We have a use case where we want to add secondary actions (or content) to the expansion panel header. This required some refactoring of the button element, which included adding an `<IconButton` so that we could allot some space for content to go between the clickable title and plus icon.

expansion-panel-secondary-header


MMFane:
I'm curious for context: could you describe the use case?

MMFane (approved):
LGTM, still curious about the use case but that shouldn't block you moving forward

dexterca:
> I'm curious for context: could you describe the use case?

@MMFane Here's an example of the use case. I'm currently accomplishing this with an absolute positioning workaround, which is not ideal. This PR allows me to remove the hack.

genomics-review-expansion-panels

PR #416, opened 2024-06-17
feat: Update chromicons and icon typings

hewayman:
The icon typings here need to be updated so the latest version of `chromicons` can be used in PHC-UI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     }
   },
   "peerDependencies": {
-    "@lifeomic/chromicons": ">=1.2.1",
+    "@lifeomic/chromicons": ">=3.0.0",
     "react": "16",
     "react-dom": "16",
     "react-router-dom": ">=5.1.2"
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "@babel/preset-env": "^7.22.10",
     "@babel/preset-react": "^7.22.5",
     "@babel/preset-typescript": "^7.22.5",
-    "@lifeomic/chromicons": "^2.2.0",
+    "@lifeomic/chromicons": "^3.4.0",
     "@storybook/addon-a11y": "^7.3.2",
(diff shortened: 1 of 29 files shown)


swain (commented):
Is this a breaking change to the types?

hewayman:
> Is this a breaking change to the types?

@swain I was trying to avoid breaking anything. This change reflects the correct types in the newest version of `chromicons`. I can't update `chromicons` in PHC-UI without breaking anything (because of the mismatched types), so I was updating the types here first. I thought the changes here would be released as a new version, so then I could use the newer versions of `chromicons` and `chroma-react` at the same time in PHC-UI. Is there a better way to handle this?

swain:
@hewayman Your logic all makes sense to me. Only reason I asked: if this _is_ a breaking change, we should mark it as such with a major version bump.

You can do that by adding a _new_ commit (or amending your existing one) to match this message syntax:

`feat!: blah blah blah`

The `!` marks the commit as a breaking change, and will cause `semantic-release` to release this change as a major version bump.

sternetj (commented):
Swain means a breaking change as in the API is changing which means when consumers of this library upgrade their version of chroma-react, they will need to make some change.

I think we should consider this breaking. I think we should update the peer dependencies to reference the lowest version of chromicons that had the API change, [v3.0.0](https://github.com/lifeomic/chromicons/tree/v3.0.0).

A "breaking change" really just means we need to bump the major version of the library. This can be done by commenting with an exclamation mark in the commit after the semmantic tag. i.e. `feat!: Update the peer dependency for chromicons`

hewayman:
@swain @sternetj That makes sense. I missed the note about `!` in the docs. I'll do that and update the peer dependencies to 3.0.0. Thanks

sternetj (approved):

